PR #11, opened 2021-12-31
Move FaissHookFvec/distances_simd/BitsetView into knowhere/utils

cydrain:


knowhere/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@
 include_directories(${INDEX_SOURCE_DIR}/knowhere)
 include_directories(${INDEX_SOURCE_DIR}/thirdparty)
 include_directories(${INDEX_SOURCE_DIR}/thirdparty/NGT/lib)
+include_directories(${INDEX_SOURCE_DIR}/knowhere/knowhere/utils)
```
jaime0815, line 17: we already have a utils dir in ${INDEX_SOURCE_DIR}/knowhere， so suggest moving these files into this dir

jaime0815: should have one utils dir for knowhere

cydrain: the file Error.h/Json.h/Status.cpp/Status.h under knowhere/utils should be moved to knowhere/knowhere/common, I'll do that using another PR

PR #16, opened 2022-01-04
Fix unittest panic

cydrain:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,18 +104,6 @@ if (MILVUS_SUPPORT_SPTAG)
     add_compile_definitions("MILVUS_SUPPORT_SPTAG")
 endif ()
 
-if (MACOS)
-    add_subdirectory(thirdparty)
-    add_subdirectory(selfdev/bruteforce)
-endif ()
-
-if (LINUX)
-    include(ThirdPartyPackagesCore)
-    add_subdirectory(thirdparty)
-endif ()
-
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -128,6 +127,10 @@ else ()
     endif ()
 endif ()
 
+if (LINUX)
```
jaime0815, line 130: delete 130-133 lines  and add "add_subdirectory(knowhere)" to 117 line.

jaime0815: both apple and linux os need to compile knowhere

CMakeLists.txt
```diff
@@ -105,16 +105,15 @@ if (MILVUS_SUPPORT_SPTAG)
...
 
 if (MACOS)
-    add_subdirectory(thirdparty)
     add_subdirectory(selfdev/bruteforce)
 endif ()
 
 if (LINUX)
     include(ThirdPartyPackagesCore)
-    add_subdirectory(thirdparty)
+    add_subdirectory(knowhere/knowhere/utils)
```
jaime0815, line 113: Can we compile utils like nsg? as sub dir of knowhere to execute add_subdirectory is weird?

cydrain: utils is not like nsg.
Because knowhere depends on faiss, and faiss depends on utils, we MUST build utils firstly, then faiss, then knowhere.

PR #180, opened 2022-05-16
Fix GPU index size calculation

cydrain:

### DIFF
--- a/knowhere/index/vector_index/gpu/IndexGPUIVF.cpp
+++ b/knowhere/index/vector_index/gpu/IndexGPUIVF.cpp
@@ -79,6 +79,15 @@ GPUIVF::CopyGpuToGpu(const int64_t device_id, const Config& config) {
     return std::static_pointer_cast<IVF>(host_index)->CopyCpuToGpu(device_id, config);
 }
 
+int64_t
+GPUIVF::Size() {
+    if (!index_) {
+        KNOWHERE_THROW_MSG("index not initialize");
+    }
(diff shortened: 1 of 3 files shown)


knowhere/index/vector_index/gpu/IndexGPUIVF.cpp
```diff
@@ -79,6 +79,15 @@ GPUIVF::CopyGpuToGpu(const int64_t device_id, const Config& config) {
     return std::static_pointer_cast<IVF>(host_index)->CopyCpuToGpu(device_id, config);
 }
 
+int64_t
+GPUIVF::Size() {
+    if (!index_) {
+        KNOWHERE_THROW_MSG("index not initialize");
```
liliuleo, line 85: Can we avoid new exception throw? Just a suggestion, since I don't have the big pic, plz feel free to ignore if it doesn't make sense.

cydrain: We need discuss this in the next release.
By now, since every index member APIs will check index initialized or not, and throw an exception, just keep this behavior sync with other APIs.

liliuleo: Sounds good to me, and thanks for the explanation.

cydrain: We will discuss this in the next release.
By now, since every index member APIs will check index initialized or not, and throw an exception, just keep it sync with other APIs.

PR #183, opened 2022-05-17
Optimize std::string usage in Dataset

cydrain:


cydrain:
After discussion, it's not safe enough to use sting_view as map key, so close this PR.

/close

Issue #231, opened 2022-07-05
Add python range search api.


Presburger:
python api support range search.

PR #232, opened 2022-07-05
Add range search python3 api and opted swig wrapper.

Presburger:
#231
Signed-off-by: Yusheng.Ma <Yusheng.Ma@zilliz.com>

### DIFF
--- a/python/knowhere/__init__.py
+++ b/python/knowhere/__init__.py
@@ -1,4 +1,5 @@
 from .swigknowhere import *
+import numpy as np
 
 
 def CreateIndex(index_name):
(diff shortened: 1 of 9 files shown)


python/knowhere/__init__.py
```diff
@@ -18,7 +23,7 @@ def CreateIndex(index_name):
         return GPUIVFSQ(-1)
     raise ValueError(
         """ index name only support 
-            'annoy' 'ivf' 'ivfsq' 'hnsw'
+            'annoy' 'ivf' 'ivfsq' 'hnsw' 'idmap' 'binary_idmap'
```
cydrain, line 26: why no "ivfpq"

PR #240, opened 2022-07-07
Add API BruteForceSearch()

cydrain:


landyliu (commented):
if this API is provided for segcore,  better put into a top direcotry,  for example , in index directory?

cydrain:
> if this API is provided for segcore, better put into a top direcotry, for example , in index directory?

Each file under index directory is serving for a single index.
For this file BruteForce.cpp, it's not for a single index, it's a collection of faiss searching operations,
we cannot put it under index directory.

PR #243, opened 2022-07-08
Add API BruteForceRangeSearch()

cydrain:


xiaofan-luan:
question, why
1.  bruteforce has to be a interface?
2. even range search, why do we have a specific interface

PR #249, opened 2022-07-11
add test-coverage flag to knowhere and remove useless directory in run_codecov.sh

foxspy:
…run_codecov.sh

Signed-off-by: xianliang <xianliang.li@zilliz.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,18 +240,19 @@ config_summary()
 
 
 #***************************** Source files ****************************
-add_subdirectory( knowhere/utils )
-add_subdirectory( thirdparty )
-add_subdirectory( knowhere )
-
-if ( KNOWHERE_BUILD_TESTS )
-    if ( BUILD_COVERAGE STREQUAL "ON" )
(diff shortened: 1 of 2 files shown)


jennyli-z:
rebase master if you want to run test

CMakeLists.txt
```diff
@@ -240,18 +240,19 @@ config_summary()
...
 
 #***************************** Source files ****************************
-add_subdirectory( knowhere/utils )
-add_subdirectory( thirdparty )
-add_subdirectory( knowhere )
-
-if ( KNOWHERE_BUILD_TESTS )
-    if ( BUILD_COVERAGE STREQUAL "ON" )
+if ( BUILD_COVERAGE STREQUAL "ON" )
         append_flags( CMAKE_CXX_FLAGS
```
cydrain, line 244: this flag should be appended only when both "BUILD_COVERAGE" and "KNOWHERE_BUILD_TESTS" enabled

foxspy: "BUILD_COVERAGE" should be isolated with unittest because other interface such as python tests also need the code coverage statistics

PR #298, opened 2022-07-21
Support DiskANN in knowhere

liliu-z:

## What Included
This PR is about adding `DiskANN` to `Knowhere`.
## Not Included
The `Bitset_view` support and `Filemanager` support will be in the following PRs.
Also some UT will follow.
## About Config
I decided to manager config individually from the `IndexParameter` + `ConfAdapter` due to:
1. `DiskANN` has lots of params and comments, that can be hard to read, manage and use in the former way.
2. `ConfAdapter` only support `Train` and `Search` check in the parent adapter class, and `DiskANN` need `Prepare` and `QueryByRange` check. With `ConfAdapter` we need to do cast in `Milvus`.
3. We merge param check and method call together to avoid misusing.


knowhere/index/vector_index/IndexDiskANN.cpp
```diff
@@ -0,0 +1,269 @@
...
+ *
+ * @return false if any error.
+ */
+auto
+ConvertId(const uint64_t* src, int64_t* des, size_t num) -> bool {
+    int64_t max = std::numeric_limits<int64_t>::max();
+    bool successful = true;
+#pragma omp parallel for schedule(static, 65536)
+    for (size_t n = 0; n < num; ++n) {
+        if (src[n] > max) {
```
cydrain, line 73: this compare will convert src[n] to int64_t implicitly, so it will always be false

knowhere/index/vector_index/IndexDiskANN.cpp
```diff
@@ -0,0 +1,269 @@
...
+    std::stringstream stream;
+    stream << build_conf.max_degree << " " << build_conf.search_list_size << " " << build_conf.search_dram_budget_gb
+           << " " << build_conf.build_dram_budget_gb << " " << build_conf.num_threads << " "
+           << build_conf.pq_disk_bytes;
+    diskann::build_disk_index<T>(data_set->Get<std::string>(meta::DATA_PATH).c_str(), index_prefix_, stream.str(),
+                                 metric_);
+}
+
+template <typename T>
+auto
```
cydrain, line 98: why not use bool return type directly ?

knowhere/index/vector_index/IndexDiskANN.cpp
liliu-z: I tried this locally
```
int main() {
        int64_t max =  numeric_limits<int64_t>::max();
        uint64_t* src = static_cast<uint64_t*>(malloc(sizeof(uint64_t) ));
        *src = 9223372036854775809;
        if (src[0] > max) {
                cout << "yes" << endl;
        }

}
```

And it printed "yes"

knowhere/index/vector_index/IndexDiskANNConfig.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+    config.at(key).get_to(to_be_set);
+}
+
+/**
+ * @brief Check the numeric param's existence and type, and allocate it to the config.
+ *
+ */
+void
+CheckBoolPrarmAndSet(const Config& config, const std::string& key, bool& to_be_set) {
```
cydrain: CheckBoolParamAndSet

knowhere/index/vector_index/IndexDiskANNConfig.cpp
```diff
@@ -0,0 +1,201 @@
...
+ * @brief Check the numeric param's existence and type, and allocate it to the config.
+ *
+ * @tparam T can only be float or int.
+ * @param min_o the min value of the value range, should be std::nullopt if we want to use numeric_limits::min().
+ * @param max_o the max value of the value range, should be std::nullopt if we want to use numeric_limits::max().
+ * @throw KnowhereException if any error.
+ */
+template <typename T>
+void
+CheckNumericPararmAndSet(const Config& config, const std::string& key, std::optional<T> min_o, std::optional<T> max_o,
```
cydrain, line 56: CheckNumericParamAndSet

knowhere/index/vector_index/IndexDiskANNConfig.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+    config.at(key).get_to(to_be_set);
+}
+
+/**
+ * @brief Check the numeric param's existence and type, and allocate it to the config.
+ *
+ */
+void
+CheckBoolParmAndSet(const Config& config, const std::string& key, bool& to_be_set) {
```
cydrain, line 75: CheckBoolParamAndSet

knowhere/index/vector_index/helpers/IndexParameter.h
```diff
@@ -34,6 +34,7 @@ constexpr const char* TOPK = "k";
 constexpr const char* RADIUS = "radius";
 constexpr const char* INPUT_IDS = "input_ids";
 constexpr const char* DEVICE_ID = "gpu_id";
+constexpr const char* DATA_PATH = "data_path";
```
cydrain, line 37: I think "DATA_PATH" should put into IndexParam

liliu-z: Since I use it in `DatasetPtr`, and noticed all `DatasetPtr` keys is in the `meta` namespace. I kept this convention in case there is any hidden meanings. Plz let me know if no, and I will put it in to the `indexparam` namespace.

knowhere/index/vector_index/IndexDiskANN.cpp
```diff
@@ -0,0 +1,269 @@
...
+    CheckPreparation(is_prepared_);
+
+    auto query_conf = DiskANNQueryConfig::Get(config);
+    auto& k = query_conf.k;
+
+    GET_TENSOR_DATA_DIM(dataset_ptr);
+    auto query = static_cast<T*>(p_data);
+    auto p_id_u64 = std::vector<uint64_t>(k * rows).data();
+    auto p_id = static_cast<int64_t*>(malloc(sizeof(int64_t) * k * rows));
+    auto p_dist = static_cast<float*>(malloc(sizeof(float) * k * rows));
```
matchyc, line 196: Would be better to use new and delete[] or simply vector to avoid deallocating resources manually?

knowhere/index/VecIndex.h
```diff
@@ -43,11 +43,26 @@ class VecIndex : public Index {
...
+        KNOWHERE_THROW_MSG("Prepare not supported yet");
+    }
+
     virtual DatasetPtr
     GetVectorById(const DatasetPtr& dataset, const Config& config) {
         KNOWHERE_THROW_MSG("GetVectorById not supported yet");
     }
 
+    /**
+     * @brief TopK Query. Will fill the return ids&distances with -1 if the result size is smaller than K.
```
cydrain, line 63: when no enough result, id will be set -1, distance will be set float_min(IP) or float_max(L2)

knowhere/index/vector_index/IndexDiskANN.cpp
```diff
@@ -0,0 +1,269 @@
...
+        metric_ = diskann::INNER_PRODUCT;
+    } else {
+        KNOWHERE_THROW_MSG("DiskANN only support L2 and IP distance.");
+    }
+}
+
+namespace {
+void
+CheckPreparation(bool is_prepared) {
+    if (is_prepared) {
```
cydrain, line 56: is this logic correct ?

knowhere/index/vector_index/IndexDiskANN.cpp
cydrain: > Would be better to use new and delete[] or simply vector to avoid deallocating resources manually?

Must use malloc here, because free is used when Dataset destruct

knowhere/index/vector_index/IndexDiskANNConfig.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+    if (!config[key].is_boolean()) {
+        KNOWHERE_THROW_FORMAT("Param '%s' should be a bool", key.data());
+    }
+    config.at(key).get_to(to_be_set);
+}
+}  // namespace
+
+void
+to_json(Config& config, const DiskANNBuildConfig& build_conf) {
```
cydrain, line 87: why not define this API like following, it's easy to use.
Config
to_json(const DiskANNBuildConfig& build_conf)

knowhere/index/vector_index/helpers/IndexParameter.h
cydrain: The data in Dataset is vector related, data in Config is index related.
"DATA_PATH" is index related, not vector related, so I think it should be in IndexParam and passed in via Config

knowhere/index/vector_index/IndexDiskANN.cpp
cydrain: this way is not safe in all platforms, better use this:
        if (src[n] && (1 << 63)) {

knowhere/index/vector_index/IndexDiskANN.cpp
liliu-z: good catch! I am working on the UT to cover these noob problems

knowhere/index/vector_index/IndexDiskANNConfig.cpp
liliu-z: This is not the API we define. The `to_json` and `from_json` are default signatures `nlohmann_json` use to serialize/deserialize object.

knowhere/index/vector_index/IndexDiskANNConfig.cpp
```diff
@@ -0,0 +1,215 @@
...
+static constexpr const char* kNumThreads = "num_threads";
+static constexpr const char* kPqDiskBytes = "pq_disk_bytes";
+
+static constexpr const char* kNumNodesToCache = "num_nodes_to_cache";
+static constexpr const char* kWarmUp = "warm_up";
+static constexpr const char* kUseBfsCache = "use_bfs_cache";
+
+static constexpr const char* kK = "k";
+static constexpr const char* kBeamwidth = "beamwidth";
+static constexpr const char* kCandidatesLimit = "candidates_limit";
```
matchyc, line 35: Does it limit anything? I saw a reference in `full_retset.reserve(candidates_limit)`, but it seems just a preallocate.

liliu-z: Good catch! I missed that. will delete it.

PR #309, opened 2022-07-23
Support Filemanager in DiskANN code

liliu-z:

## General
This PR did two things:
1. We supported `FileManager` in `DiskANN`.
2. We collected all filenames that need to be managed by the `FileManager` and wrote getter inside `DiskANN` source code to get them. Since `DiskANN` now manually form the filename whenever it need them, which is super error-prone.

## FileManager
This is the first step of the `FileManager`. In the plan, we have two steps in total:
1. `FileManager` will help in file granularity. We call `AddFile` to register file to the `FileManager` and call `LoadFile` to ensure `FileManager` prepare it well to use. This way is clean and easy to implement. The drawbacks are:
  a. `Knowhere` should know the existence of a remote platform, so it needs to proactively call method to prepare the files.
  b.  If the file is too large to fit the Disk, this method cannot work.
2. In this step we plan to solve the problems mentioned before. We need to use `FileManager` as a `ifstream` + `ofstream`, so it can support finer granularity.

### DIFF
--- a/thirdparty/DiskANN/include/pq_table.h
+++ b/thirdparty/DiskANN/include/pq_table.h
@@ -45,11 +45,12 @@ namespace diskann {
 #else
     void load_pq_centroid_bin(const char* pq_table_file, size_t num_chunks) {
 #endif
-        std::string rearrangement_file = std::string(pq_table_file) +
-                                         "_rearrangement_perm.bin";
+        std::string rearrangement_file =
+            get_pq_rearrangement_perm_filename(std::string(pq_table_file));
     std::string chunk_offset_file =
-        std::string(pq_table_file) + "_chunk_offsets.bin";
-    std::string centroid_file = std::string(pq_table_file) + "_centroid.bin";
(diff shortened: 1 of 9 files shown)


thirdparty/DiskANN/include/pq_table.h
```diff
@@ -110,7 +111,7 @@ namespace diskann {
         rearrangement[d] = d;
       chunk_offsets = new uint32_t[num_chunks + 1];
       for (uint32_t d = 0; d <= num_chunks; d++)
-        chunk_offsets[d] = (_u32)(std::min)(ndims, d * chunk_size);
+        chunk_offsets[d] = (_u32) (std::min)(ndims, d * chunk_size);
```
cydrain, line 114: I think this line will break lint check

liliu-z: From the CI, the lint is fine. But we should not change the original format of the 3rd party code. This must be accidentally changed by VSCode formatting, will change it back.

PR #316, opened 2022-07-26
Set correct metric type for RHNSW_PQ

cydrain:
Issue: https://github.com/milvus-io/milvus/issues/18361
issue：#18361


landyliu:
Need to add UT

cydrain:
> Need to add UT

I'll add IP unittest for all index types using another PR

PR #317, opened 2022-07-26
Fix some problem in DiskANN support

liliu-z:

Fix some problems in previous implementation:
1. Add `IndexDiskANNConfig` to CMakeLists
2. Mark `FileManager` pure virtual class
3. Modify the param check. Since we cannot include .cpp file, so we cannot reuse the check method in `AdapterConf.cpp`.
4. Fix some other bugs

### DIFF
--- a/knowhere/index/vector_index/IndexDiskANN.h
+++ b/knowhere/index/vector_index/IndexDiskANN.h
@@ -57,9 +57,6 @@ class IndexDiskANN : public VecIndex {
     AddWithoutIds(const DatasetPtr& data_set, const Config& config) override;
 
     bool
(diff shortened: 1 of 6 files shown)


knowhere/index/vector_index/IndexDiskANN.h
```diff
@@ -56,9 +56,6 @@ class IndexDiskANN : public VecIndex {
     void
     AddWithoutIds(const DatasetPtr& data_set, const Config& config) override;
 
-    bool
-    BuildDiskIndex(const Config& config) override;
```
cydrain, line 60: BuildIndex() is better ?

PR #321, opened 2022-07-27
Reduce the memory usage of thread data's coord_scratch

cqy123456:
**There are two main reasons for optimizing coord_scratch:**
1.  DiskANN applies for a large amount of memory space for coord_scratch.  But coord_scratch only uses a small amount of space to store calculated data；
2. DiskANN uses the wrong indexing mode (i * dim), the correct one is (i* dim * sizeof(T)). Therefore, it is easy to cause out-of-bounds access.

**Therefore, this optimization achieves the following two points:**
1.  Allocate as much memory space to coord_scratch as the search uses;
2. The data to be calculated is directly copied to the starting position of 

**The following is the experimental analysis of sift1M:**

-    Experiment environment:  AWS m6id.2xlarge;

-    Preparation :  No cache, Threads number = 8;

image

After optimization, DiskANN can reduce a lot of thread data memory space. It can be observed that the memory footprint of thread data is negligible compared to the PQ compression code.

### DIFF
--- a/thirdparty/DiskANN/include/pq_flash_index.h
+++ b/thirdparty/DiskANN/include/pq_flash_index.h
@@ -21,7 +21,6 @@
 #include "windows_customizations.h"
 
 #define MAX_GRAPH_DEGREE 512
-#define MAX_N_CMPS 16384
 #define SECTOR_LEN (_u64) 4096
 #define MAX_N_SECTOR_READS 128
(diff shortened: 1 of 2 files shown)


liliu-z:
Plz describe the reason we do this optimization and why it work. Also attach the experiment result here for recording purpose.
And one PR need to have one commit, plz squash these commits and force push.

PR #320, opened 2022-07-27
Link MKL to libknowhere

matchyc:
Signed-off-by: matchyc <dawnlight.yc@protonmail.com>

This link action will be triggered when building Knowhere with DiskANN, otherwise, it won't happen.

PR #333, opened 2022-07-28
Avoid using find_package(MKL) to fix error

matchyc:
Signed-off-by: matchyc <dawnlight.yc@protonmail.com>

find_package(MKL) has different behaviors in Ubuntu 18.04/20.04/22.04, which results in errors in 22.04. So we avoid using it to ensure it works in the three Ubuntu versions.

related: #320


matchyc:
@Presburger

PR #337, opened 2022-08-01
Fix: query vector copy out of bounds when using InnnerProduct

SonglinLife:
Signed-off-by: wusonglin <wusonglin@tongji.edu.cn>

issue: #253 

diskann uses norm L2 to replace InnerProduct when using SSD index. It adds an extra dim at the end, this would cause a  array out of bounds issue.
image

### DIFF
--- a/thirdparty/DiskANN/src/pq_flash_index.cpp
+++ b/thirdparty/DiskANN/src/pq_flash_index.cpp
@@ -11,6 +11,7 @@
 #include <chrono>
 #include <cmath>
 #include <iterator>
+#include <random>
 #include <thread>
 #include "distance.h"
 #include "exceptions.h"
@@ -842,7 +843,12 @@ namespace diskann {
     const T *    query = data.scratch.aligned_query_T;
     const float *query_float = data.scratch.aligned_query_float;
 
-    for (uint32_t i = 0; i < this->data_dim; i++) {
+    auto q_dim = this->data_dim;
+    if(metric== diskann::Metric::INNER_PRODUCT ){
+      // query_dim need to be specially treated when using IP
+      q_dim--;
+    }
+    for (uint32_t i = 0; i < q_dim; i++) {
       data.scratch.aligned_query_float[i] = query1[i];
       data.scratch.aligned_query_T[i] = query1[i];
       query_norm += query1[i] * query1[i];
@@ -851,6 +857,12 @@ namespace diskann {
     // if inner product, we laso normalize the query and set the last coordinate
     // to 0 (this is the extra coordindate used to convert MIPS to L2 search)
     if (metric == diskann::Metric::INNER_PRODUCT) {
+      if(query_norm == 0){
+        // return an empty answer when calcu a zero point 
+        this->thread_data.push(data);
+        this->thread_data.push_notify_all();
+        return;
+      }
       query_norm = std::sqrt(query_norm);
       data.scratch.aligned_query_T[this->data_dim - 1] = 0;
       data.scratch.aligned_query_float[this->data_dim - 1] = 0;
--- a/thirdparty/DiskANN/src/aux_utils.cpp
+++ b/thirdparty/DiskANN/src/aux_utils.cpp
@@ -1124,7 +1124,7 @@ namespace diskann {
                                    ? MAX_SAMPLE_POINTS_FOR_WARMUP
                                    : ten_percent_points;
     double sample_sampling_rate = num_sample_points / points_num;
-    gen_random_slice<T>(data_file_to_use.c_str(), sample_data_file,
+    gen_random_slice<T>(base_file.c_str(), sample_data_file,
                         sample_sampling_rate);
 
     if (compareMetric == diskann::Metric::INNER_PRODUCT) {


thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -842,12 +843,33 @@ namespace diskann {
...
+    if(metric== diskann::Metric::INNER_PRODUCT && !smaple){ // query_dim need specify when using IP
+      q_dim--;
+    }
+    for (uint32_t i = 0; i < q_dim; i++) {
       data.scratch.aligned_query_float[i] = query1[i];
       data.scratch.aligned_query_T[i] = query1[i];
       query_norm += query1[i] * query1[i];
     }
 
+    if(query_norm <= 0 && metric == diskann::Metric::INNER_PRODUCT){
```
liliu-z, line 856: To my understanding, if query_norm == 0, DiskANN will crash right? And did we meet this situation before?

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -842,12 +843,33 @@ namespace diskann {
...
+    // random answer for zero point ip search
```
liliu-z, line 857: Why we do random answer

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -842,12 +843,33 @@ namespace diskann {
...
+        }
+      }
+      this->thread_data.push(data);
+      this->thread_data.push_notify_all();
+      return;
+    }
+
     // if inner product, we laso normalize the query and set the last coordinate
     // to 0 (this is the extra coordindate used to convert MIPS to L2 search)
     if (metric == diskann::Metric::INNER_PRODUCT) {
```
liliu-z, line 875: Can you merge the query_norm check with this?

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -825,7 +826,7 @@ namespace diskann {
                                            const _u64  beam_width,
                                            const bool  use_reorder_data,
                                            QueryStats *stats,
-                                           faiss::BitsetView bitset_view) {
+                                           faiss::BitsetView bitset_view, bool smaple=false) {
```
liliu-z, line 829: Typo. and can you explain this flag? Also the default value should be in .h file right?

liliu-z (changes requested):
Plz leave some comments to explain this change in detail and give some test evidence for recording purpose.

thirdparty/DiskANN/src/pq_flash_index.cpp
SonglinLife: No, it wouldn't crash, because query1's mem allocated at heap. if access the out-of-bounds mem, it would get a random value.

SonglinLife: > To my understanding, if query_norm == 0, DiskANN will crash right? And did we meet this situation before?

No, it wouldn't crash, because query1's mem allocated at heap. if access the out-of-bounds mem, it would get a random value.

thirdparty/DiskANN/src/pq_flash_index.cpp
SonglinLife: > Why we do random answer

The distance between zero point and all points is the same, it is better to return a random answer

thirdparty/DiskANN/src/pq_flash_index.cpp
SonglinLife: > Typo. and can you explain this flag? Also the default value should be in .h file right?

sorry forgot this :)

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -842,15 +845,27 @@ namespace diskann {
...
       data.scratch.aligned_query_T[i] = query1[i];
       query_norm += query1[i] * query1[i];
     }
 
+
     // if inner product, we laso normalize the query and set the last coordinate
     // to 0 (this is the extra coordindate used to convert MIPS to L2 search)
     if (metric == diskann::Metric::INNER_PRODUCT) {
+      if(query_norm == 0){
+        // return an empty answer when calcu a zero point
```
liliu-z, line 864: Since query on 0 vector is illegal, do you think throw an exception is a better way to do?

thirdparty/DiskANN/src/pq_flash_index.cpp
liliu-z: Interesting, did you see this? the number should be extremely big and the normalization will be super weird.

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -842,15 +845,27 @@ namespace diskann {
...
+      // query_dim need to be specially treated when using IP
+      q_dim--;
+    }
+    for (uint32_t i = 0; i < q_dim; i++) {
       data.scratch.aligned_query_float[i] = query1[i];
       data.scratch.aligned_query_T[i] = query1[i];
       query_norm += query1[i] * query1[i];
     }
 
+
```
liliu-z, line 859: Plz do not add empty line.

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -321,12 +322,14 @@ namespace diskann {
     std::vector<uint64_t> tmp_result_ids_64(sample_num, 0);
     std::vector<float>    tmp_result_dists(sample_num, 0);
 
+    this->use_sample = true;
```
liliu-z, line 325: Can you explain why we need this?

Also I don't think it is a good idea to add it to the param list of `cached_beam_search` since the API need to be as lite as possible. This one can be a field of the class since no one outside need to set it right?

thirdparty/DiskANN/src/pq_flash_index.cpp
SonglinLife: > Since query on 0 vector is illegal, do you think throw an exception is a better way to do?

No, in my opinion, zero point has its physical significance. see it 🗯️  [zero vector](https://qr.ae/pvuAh1). Returning an empty answer or a random answer list is both ok, but throwing an exception maybe not be a good idea, because zero vector can be used in mathematical calculations.
when it comes to a **NULL** vector, it should throw an exception.

thirdparty/DiskANN/src/pq_flash_index.cpp
SonglinLife: > Can you explain why we need this?
> 
> Also I don't think it is a good idea to add it to the param list of `cached_beam_search` since the API need to be as lite as possible. This one can be a field of the class since no one outside need to set it right?

became when diskann generates cache nodes, it used sample vector data in the build stage. when using IP metric, diskann will add an extra dim, like 256(original dim) -> 257(L2 norm add a dim) -> 264(aligned dim). but query vector data have not been L2 normed, it's dim like 256 -> 256 (aligned dim).

yes, I move the sample field from `cache_beam_search`, and use an extra flag to indicate generate cache stage. the same idea also see  [diskann](https://github.com/microsoft/DiskANN/blob/main/src/pq_flash_index.cpp#L293)

liliu-z: Do you mind to add comment about this here. Or it is hard to understand

PR #352, opened 2022-08-03
Catch diskann exceptions

liliu-z:

issue: #253 

This PR is to catch all exceptions from `DiskANN` in `Knowhere`. By doing this Knowhere get the full control of error handling.

### DIFF
--- a/knowhere/index/vector_index/IndexDiskANN.cpp
+++ b/knowhere/index/vector_index/IndexDiskANN.cpp
@@ -98,8 +98,8 @@ CheckFileSize(const std::string& data_path, const size_t num, const size_t dim) 
     file.close();
     uint64_t expected_file_size = num * dim * sizeof(T) + 2 * sizeof(uint32_t);
     if (autual_file_size != expected_file_size) {
-        KNOWHERE_THROW_FORMAT("Actual file size (%ld bytes) not equal to expected size (%ld bytes)",
-                              autual_file_size, expected_file_size);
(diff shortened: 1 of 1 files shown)


knowhere/index/vector_index/IndexDiskANN.cpp
```diff
@@ -319,6 +435,10 @@ IndexDiskANN<T>::QueryByRange(const DatasetPtr& dataset_ptr, const Config& confi
         *(p_lims + row + 1) = res_count;
     }
 
+    if (ex.has_value()) {
+        throw ex.value();
```
cydrain, line 439: why not use KNOWHERE_THROW_MSG ?

knowhere/index/vector_index/IndexDiskANN.cpp
```diff
@@ -277,10 +362,28 @@ IndexDiskANN<T>::Query(const DatasetPtr& dataset_ptr, const Config& config, cons
...
+            {
+                if (ex == std::nullopt) {
+                    ex = std::make_optional<KnowhereException>(e);
+                }
+            }
+        }
+    }
+
+    if (ex.has_value()) {
+        throw ex.value();
```
cydrain, line 386: should use KNOWHERE_THROW_MSG

knowhere/index/vector_index/IndexDiskANN.cpp
liliu-z: ex is `optional<KnowhereException>` which we get from `KNOWHERER_THROW_MSG`. Since we cannot throw exceptions from openmp threads, so we collect them and do rethrow here

PR #355, opened 2022-08-04
Add unittest for build_config check function

cqy123456:

issue: #253 

This unittest uses unreasonable build index params to check if knowhere::KnowhereException can be thrown when data size != file size or diskann's node size > a sector.

### DIFF
--- a/unittest/test_diskann.cpp
+++ b/unittest/test_diskann.cpp
@@ -150,11 +150,11 @@ GenRangeSearchGrounTruth(const float* data_p, const float* query_p, const std::s
 }
 
 void
-WriteRawDataToDisk(const std::string data_path, const float* raw_data) {
+WriteRawDataToDisk(const std::string data_path, const float* raw_data, const uint32_t num, const uint32_t dim) {
     std::ofstream writer(data_path.c_str(), std::ios::binary);
-    writer.write((char*)&kNumRows, sizeof(uint32_t));
-    writer.write((char*)&kDim, sizeof(uint32_t));
-    writer.write((char*)raw_data, sizeof(float) * kNumRows * kDim);
+    writer.write((char*)&num, sizeof(uint32_t));
+    writer.write((char*)&dim, sizeof(uint32_t));
+    writer.write((char*)raw_data, sizeof(float) * num * dim);
     writer.close();
 }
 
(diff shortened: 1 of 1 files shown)


unittest/test_diskann.cpp
```diff
@@ -423,4 +423,58 @@ TEST_P(DiskANNTest, config_test) {
...
+        std::vector<float> raw_data(data_dim * data_n);
+        WriteRawDataToDisk(test_data_path, raw_data.data(), data_n, data_dim);
+
+        illegal_build_config = build_conf; // init
+        illegal_build_config.data_path = test_data_path;
+        illegal_build_config.max_degree = std::get<1>(dim_and_R);
+        knowhere::DiskANNBuildConfig::Set(illegal_cfg, illegal_build_config);
+        knowhere::IndexDiskANN<float> test_diskann(test_index_path, metric_,
+                                                   std::make_unique<knowhere::LocalFileManager>());
+        ASSERT_THROW(test_diskann.BuildAll(nullptr, illegal_cfg), knowhere::KnowhereException);
```
liliu-z, line 456: Why don't use expect throw

unittest/test_diskann.cpp
```diff
@@ -423,4 +423,58 @@ TEST_P(DiskANNTest, config_test) {
 
     range_search_conf_to_test.beamwidth = 129;
     check_config_error<knowhere::DiskANNQueryByRangeConfig>(range_search_conf_to_test);
+}
+
+TEST_P(DiskANNTest, build_config_test) {
+    std::string test_data_path = kDir + "/diskann_build_config_test_data";
```
liliu-z, line 429: Can we create a folder for this test?

unittest/test_diskann.cpp
```diff
@@ -423,4 +423,62 @@ TEST_P(DiskANNTest, config_test) {
...
+
+        illegal_build_config = build_conf; // init
+        illegal_build_config.data_path = test_data_path;
+        illegal_cfg.clear();
+        knowhere::DiskANNBuildConfig::Set(illegal_cfg, illegal_build_config);
+        knowhere::IndexDiskANN<float> test_diskann(test_index_dir, metric_,
+                                                   std::make_unique<knowhere::LocalFileManager>());
+        EXPECT_THROW(test_diskann.BuildAll(nullptr, illegal_cfg), knowhere::KnowhereException);
+    }
+    fs::remove_all(test_dir);
```
liliu-z, line 483: remove all will only remove data in this folder, you need to call remove after this to remove the folder.

PR #356, opened 2022-08-04
Fix error message in DiskANN

hhy3:

issue: #253 

The libaio wrapper doesn't set errno but instead returns the negation of error number to indicating the error.

### DIFF
--- a/thirdparty/DiskANN/src/linux_aligned_file_reader.cpp
+++ b/thirdparty/DiskANN/src/linux_aligned_file_reader.cpp
@@ -135,10 +135,10 @@ void LinuxAlignedFileReader::register_thread() {
   int          ret = io_setup(MAX_EVENTS, &ctx);
   if (ret != 0) {
     lk.unlock();
-    assert(errno != EAGAIN);
-    assert(errno != ENOMEM);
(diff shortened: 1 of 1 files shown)


thirdparty/DiskANN/src/linux_aligned_file_reader.cpp
```diff
@@ -135,10 +135,10 @@ void LinuxAlignedFileReader::register_thread() {
...
   if (ret != 0) {
     lk.unlock();
-    assert(errno != EAGAIN);
-    assert(errno != ENOMEM);
-    std::cerr << "io_setup() failed; returned " << ret << ", errno=" << errno
-              << ":" << ::strerror(errno) << std::endl;
+    assert(-ret != EAGAIN);
+    assert(-ret != ENOMEM);
+    std::cerr << "io_setup() failed; returned " << ret << ", errno=" << -ret
+              << ":" << ::strerror(-ret) << std::endl;
```
liliu-z, line 141: to my understanding, if any failure `aio` returns negative errno https://man7.org/linux/man-pages/man2/io_getevents.2.html

hhy3: Yes.

PR #370, opened 2022-08-10
Update interface BruteForceSearch() and BruteForceRangeSearch()

cydrain:


cydrain:
> looks good to me, just a quick question. Noticed that we are using faiss to implement the brute force solution, will we have more in the future? Or in another words, will the implementation of different libs make any difference?
> 
> I guess no. If so, this looks fine, and we only need to modify the implementation if we find better one in the future. @cydrain

If we support more metric types in the future, we will update BruteForceSearch() and BruteForceRangeSearch() to support them.
Brute force search is not related with index types.

PR #372, opened 2022-08-10
api for concurrency when build and query

smellthemoon:
Signed-off-by: xinguo.li  <xinguo.li@zilliz.com>

issue: #374 

We added two parameters in the config, build_thread_num and query_thread_num. It can help to control the number of threads if you input parameters when build and query. Only hnsw and ivy_flat were accepted the changes for the reason that we need to solve problems in a short time.


smellthemoon:
api for concurrency when build and query

PR #376, opened 2022-08-11
Fix thread problem for DiskANN

liliu-z:
Signed-off-by: liuli <li.liu@zilliz.com>

**Problem**: We need to set omp thread limit for each thread using omp block. Now we only set in `Prepare()`, because in original `DiskANN`, they do prepare and search in one thread. So we need to set this limit every time we do query.

### DIFF
--- a/knowhere/index/vector_index/IndexDiskANN.h
+++ b/knowhere/index/vector_index/IndexDiskANN.h
@@ -101,6 +101,7 @@ class IndexDiskANN : public VecIndex {
     std::unique_ptr<diskann::PQFlashIndex<T>> pq_flash_index_;
 
     bool is_prepared_ = false;
+    int32_t num_threads_  = 0;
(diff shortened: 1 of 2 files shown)


knowhere/index/vector_index/IndexDiskANN.h
```diff
@@ -101,6 +101,7 @@ class IndexDiskANN : public VecIndex {
     std::unique_ptr<diskann::PQFlashIndex<T>> pq_flash_index_;
 
     bool is_prepared_ = false;
+    int32_t num_threads_  = 0;
```
cydrain, line 104: initial value 0 is ok ?
should set to 1 ?

liliu-z: This is not a default num, we will set it in `prepare()`, and without preparation, we will throw.

PR #380, opened 2022-08-12
Fix and optimize DiskANN kmeans code

hhy3:

issue: #253 

1. Make comparison operators of `PivotContainer` more intuitive.
2. Optimize algorithm for finding k nearest centers in `kmeans::compute_closest_centers_in_block` from $O(nlogn)$ to $O(n)$.
3. Fix potential memory leak in `kmeans::run_lloyds`.
4. Fix bug of `kmeans::selecting_pivots`

### DIFF
--- a/thirdparty/DiskANN/src/math_utils.cpp
+++ b/thirdparty/DiskANN/src/math_utils.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 #include <limits>
+#include <unordered_set>
 #include <malloc.h>
 #include <math_utils.h>
 #include <mkl.h>
@@ -109,16 +110,15 @@ namespace math_utils {
     } else {
 #pragma omp parallel for schedule(static, 8192)
       for (int64_t i = 0; i < (_s64) num_points; i++) {
-        std::priority_queue<PivotContainer> top_k_queue;
+        std::vector<PivotContainer> top_k_vec;
         float* current = dist_matrix + (i * num_centers);
         for (size_t j = 0; j < num_centers; j++) {
-          PivotContainer this_piv(j, current[j]);
-          top_k_queue.push(this_piv);
+          top_k_vec.emplace_back(j, -current[j]);
         }
-        for (size_t j = 0; j < k; j++) {
-          PivotContainer this_piv = top_k_queue.top();
-          center_index[i * k + j] = (uint32_t) this_piv.piv_id;
-          top_k_queue.pop();
+        std::nth_element(top_k_vec.begin(), top_k_vec.begin() + k - 1, top_k_vec.end());
+        std::sort(top_k_vec.begin(), top_k_vec.begin() + k);
+        for (int64_t j = 0; j < k; j++) {
+          center_index[i * k + j] = top_k_vec[j].piv_id;
         }
       }
     }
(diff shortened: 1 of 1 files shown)


thirdparty/DiskANN/src/math_utils.cpp
```diff
@@ -370,12 +381,13 @@ namespace kmeans {
...
 
     size_t tmp_pivot;
-    for (size_t j = 0; j < num_centers; j++) {
-      tmp_pivot = distribution(generator);
-      if (std::find(picked.begin(), picked.end(), tmp_pivot) != picked.end())
-        continue;
-      picked.push_back(tmp_pivot);
-      std::memcpy(pivot_data + j * dim, data + tmp_pivot * dim,
+    for (int64_t j = num_points - num_centers; j < num_points; j++) {
+      tmp_pivot = std::uniform_int_distribution<size_t>(0, j)(generator);
```
liliu-z, line 385: use set to replace vector is good, but can you explain a little bit about this change here?

liliu-z (commented):
Love these optimization. We'd better have a performance comparison attached to the fix to claim the gain.
Also, plz explain the bug more here.

thirdparty/DiskANN/include/utils.h
```diff
@@ -903,11 +903,11 @@ struct PivotContainer {
   }
 
   bool operator<(const PivotContainer& p) const {
-    return p.piv_dist < piv_dist;
```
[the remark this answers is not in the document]
hhy3: all right

thirdparty/DiskANN/src/math_utils.cpp
```diff
@@ -242,21 +242,26 @@ namespace kmeans {
...
+    bool ret_closest_docs = true;
+    if (closest_center == NULL) {
       closest_center = new uint32_t[num_points];
-    if (closest_docs == NULL)
+      ret_closest_center = false;
+    }
+    if (closest_docs == NULL) {
       closest_docs = new std::vector<size_t>[num_centers];
-    else
+      ret_closest_docs = false;
```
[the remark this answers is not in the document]
hhy3: Right, these do not happen. In addition, the comment above this function is confusing.

thirdparty/DiskANN/src/math_utils.cpp
hhy3: In the original code there's the probability that `picked.size()` is less than `num_centers` because of `continue` in the for loop.
I used an algorithm called [Bob Floyds sampling](https://math.stackexchange.com/questions/178690/whats-the-proof-of-correctness-for-robert-floyds-algorithm-for-selecting-a-sin) to sample `num_centers` points from data.

liliu-z: I didn't fully get, what will happen if `picked.size()` is less then `num_centers`?

BTW there is another place use this picked logic in this file, do we need to make them sync to each other if this change is necessary?

hhy3: If `picked.size()` is less than `num_centers`, some data of `pivot_data` will be some useless random value. It'll cause problems in the subsequent kmeans code.

That place actually used another algorithm, so there's no problem.

liliu-z: Aha, you are right. I read the wrong line in code. Good catch!

liliu-z: BTW, this algo looks reasonable to me. But what's the benefit of using this instead of simply use 
```
if (picked.count(tmp_pivot)) {
    j--;
   continue;
}
```

hhy3: There is no obvious difference in performance

PR #389, opened 2022-08-14
Fix DiskANN omp problem

hhy3:

issue: #388 

This PR removes an unnecessary critical section and gains a 2x speedup of PQ when using 16 threads.

### DIFF
--- a/thirdparty/DiskANN/src/math_utils.cpp
+++ b/thirdparty/DiskANN/src/math_utils.cpp
@@ -176,19 +176,19 @@ namespace math_utils {
           pts_norms_blk, pivs_norms_squared, closest_centers, distance_matrix,
           k);
 
+      int64_t blk_st = cur_blk * PAR_BLOCK_SIZE;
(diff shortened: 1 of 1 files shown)


thirdparty/DiskANN/src/math_utils.cpp
```diff
@@ -176,19 +176,19 @@ namespace math_utils {
...
           closest_centers_ivf[j * k + l] = (uint32_t) this_center_id;
-          if (inverted_index != NULL) {
-#pragma omp critical
-            inverted_index[this_center_id].push_back(j);
-          }
+        }
+      }
+      if (inverted_index != NULL) {
+        for (int64_t j = 0; j < num_pts_blk * k; ++j) {
+          inverted_index[closest_centers[j]].push_back(blk_st + j / k);
```
liliu-z, line 191: Do we need to use omp here?

hhy3: The upper bound of the input size of this function is about 256000. For this input size, this code segment just runs ~1ms each iteration, whereas the whole function runs ~200ms each iteration. So there is no need to use omp to optimize it.

liliu-z:
\lgtm
\approve

PR #397, opened 2022-08-16
Feat: add fast build index option

SonglinLife:
Signed-off-by: wusonglin] <songlin.wu@zilliz.com>
Since diskann has a tow-rounds build index, it is clear that only one round can save a lot of time for index building. And the recall loss is very small which can be ignored in many scenarios.
more information: 
[DiskAnn/SSD_HNSW Tests on AWS m6id.2xlarge](https://zilliverse.feishu.cn/sheets/shtcn6NrMjb1TwVBszveHyECnFd?sheet=8e0b9d) shows this optimization has saved 35% build time and lost no more than 1% recall in two KNN search dataset.
[index build optimized records](https://zilliverse.feishu.cn/docx/doxcnYeKNTJwIcxxZzcEARI2Eqg) document the optimization ideas in detail. small sync size in round 1 and more entry points help the vanama graph grows better at the Initial.

### DIFF
--- a/thirdparty/DiskANN/include/aux_utils.h
+++ b/thirdparty/DiskANN/include/aux_utils.h
@@ -99,7 +99,7 @@ namespace diskann {
   template<typename T>
   DISKANN_DLLEXPORT int build_merged_vamana_index(
       std::string base_file, diskann::Metric _compareMetric, unsigned L,
-      unsigned R, double sampling_rate, double ram_budget,
+      unsigned R, bool accelerate_build, double sampling_rate, double ram_budget,
       std::string mem_index_path, std::string medoids_file,
       std::string centroids_file);
 
--- a/thirdparty/DiskANN/src/aux_utils.cpp
+++ b/thirdparty/DiskANN/src/aux_utils.cpp
@@ -544,7 +544,7 @@ namespace diskann {
   template<typename T>
   int build_merged_vamana_index(std::string     base_file,
                                 diskann::Metric compareMetric, unsigned L,
-                                unsigned R, double sampling_rate,
+                                unsigned R, bool accelerate_build, double sampling_rate,
                                 double ram_budget, std::string mem_index_path,
                                 std::string medoids_file,
                                 std::string centroids_file) {
@@ -565,6 +565,7 @@ namespace diskann {
       paras.Set<unsigned>("num_rnds", 2);
       paras.Set<bool>("saturate_graph", 1);
       paras.Set<std::string>("save_path", mem_index_path);
+      paras.Set<bool>("accelerate_build", accelerate_build);
 
       std::unique_ptr<diskann::Index<T>> _pvamanaIndex =
           std::unique_ptr<diskann::Index<T>>(new diskann::Index<T>(
@@ -605,6 +606,7 @@ namespace diskann {
       paras.Set<unsigned>("num_rnds", 2);
       paras.Set<bool>("saturate_graph", 0);
       paras.Set<std::string>("save_path", shard_index_file);
+      paras.Set<bool>("accelerate_build", accelerate_build);
 
       _u64 shard_base_dim, shard_base_pts;
       get_bin_metadata(shard_base_file, shard_base_pts, shard_base_dim);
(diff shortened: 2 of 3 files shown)


thirdparty/DiskANN/src/index.cpp
```diff
@@ -1461,6 +1462,9 @@ namespace diskann {
     // random other nodes
     std::set<unsigned> unique_start_points;
     unique_start_points.insert(_ep);
+    if(fast)
+      while (unique_start_points.size() < 128)
+        unique_start_points.insert(_nd * dis(gen));
```
liliu-z, line 1467: Plz use braces here

thirdparty/DiskANN/src/index.cpp
```diff
@@ -1486,7 +1490,12 @@ namespace diskann {
 
       std::vector<std::vector<unsigned>> pruned_list_vector(round_size);
 
-      for (uint32_t sync_num = 0; sync_num < num_syncs; sync_num++) {
+      auto round_num_syncs = num_syncs;
+      if(fast && rnd_no == 0){
```
liliu-z, line 1494: I remember we will reduce the sync size in the second round right?

thirdparty/DiskANN/include/aux_utils.h
```diff
@@ -99,7 +99,7 @@ namespace diskann {
   template<typename T>
   DISKANN_DLLEXPORT int build_merged_vamana_index(
       std::string base_file, diskann::Metric _compareMetric, unsigned L,
-      unsigned R, double sampling_rate, double ram_budget,
+      unsigned R, bool fast, double sampling_rate, double ram_budget,
```
liliu-z, line 102: Can we name it in a more clear way? Such as `accelerate_build`, etc. PS, this is just an example.

thirdparty/DiskANN/src/aux_utils.cpp
```diff
@@ -964,6 +968,13 @@ namespace diskann {
       }
     }
 
+    bool fast = false;
+    if (param_list.size() == 8){
+      if(1 == atoi(param_list[7].c_str())){
+        fast = true;
+      }
```
liliu-z, line 975: line 953-965 of the original file also need to be taken care of.
Actually this param pass method is awful, since it brings a lot of troubles in param checking, reading, and using. If possible, can you help rephrase these param method by putting all params in the signature with default values?

thirdparty/DiskANN/src/index.cpp
SonglinLife: > I remember we will reduce the sync size in the second round right?

no, we use a very small sync size in round one with alpha 1.0. It works like adding multi-eps.

liliu-z: Cool, we need more materials in this PR:
1. explanation of how we do this acceleration.
2. test results.

thirdparty/DiskANN/src/aux_utils.cpp
SonglinLife: > line 953-965 of the original file also need to be taken care of. Actually this param pass method is awful, since it brings a lot of troubles in param checking, reading, and using. If possible, can you help rephrase these param method by putting all params in the signature with default values?

I am not sure how to improve it. Do we just modify the check params code in this function and not change any things, or also modify the params pass method which may involve UT and other files?

liliu-z: We can modify the signature of `build_disk_index` by accepting all params. Only `IndexDiskANN` will get affected.
You can either do it here, or create another PR for it and merge that first.
Also the param check logic in `build_disk_index` also need to be changed.

thirdparty/DiskANN/src/index.cpp
SonglinLife: I edit the pr comment, adding two documents. It may help better understand this pr.

thirdparty/DiskANN/src/aux_utils.cpp
SonglinLife: ok，I will modify it in this pr

Issue #402, opened 2022-08-17
IVF is not thread safe


liliu-z:
`Mivuls` issue: https://github.com/milvus-io/milvus/issues/18509
`faiss`' IVFFlat is not thread safe.

yanliang567:
what about ivf_sq8 and ivf_pq?

PR #403, opened 2022-08-17
Add thread safe search for faiss IVFFlat

liliu-z:

issue: #402 

## Background
Issue from `Milvus`: https://github.com/milvus-io/milvus/issues/18509
`IVFFlat` is not thread-safe:
1. `IVF` use `nprobe` to [init a array](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIVF.cpp#L319-L320), this `nprobe` is set by users. And then, in `search_preassigned` method, it use the [field](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIVF.cpp#L398) `nprobe` to iterate the [array](https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIVF.cpp#L539-L544). This problems also happen to `range_search`, etc.
2. this `nprobe`, and `parallel_mode` can be set in `knowhere` by a query, and get changed by another before the first query use it.

## Solution
We create a group of new function accept the search param such as `nprobe`,`parallel_mode` to make it thread safe.

## What does this PR do 
1. We create a `IndexFlat-thread_safe.cpp` file to maintain all these `knowhere` created functions. This make us easy to control and delete in the future if needed.
2. We move `search_without_codes`, `search_preassigned_without_codes`, `range_search_without_codes` and `range_search_preassigned_without_codes` to this new file. Since all these function is created by `knowhere` before. (BTW, we need to refactor these code, too many dup code which brings troubles to the future's modification). Also we rename them with `thread safe` prefix.
3. We create `thread_safe_search` and `thread_safe_range_search` in this new file.
4. We modify the IVFSearchParameters to have `parallel_mode` in it. Since we have race condition in this field too.
5. Modify `IndexIVF_NM` and `IndexIVF` in `knowhere` to support `IVFFLAT`, `IVFPQ` and `IVFSQ`.

## Function structure changes
Below shows the function call chains of `IVF`
### Before change
```
search                         ->     search_preassigned
search_without_codes           ->     search_preassigned_without_codes
range_search                   ->     range_search_preassigned
range_search_without_codes     ->     range_search_preassigned_without_codes
```
### After change
```
search/{thread_safe_search}                  ->     search_preassigned
{search_without_codes}                       ->     {search_preassigned_without_codes}
range_search/{thread_safe_range_search}      ->     range_search_preassigned
{range_search_without_codes}                 ->     {range_search_preassigned_without_codes}
```
the function name in the braces is in the new file. We didn't show the name changes here


thirdparty/faiss/faiss/IndexIVF.h
```diff
@@ -67,7 +67,9 @@ struct Level1Quantizer {
 struct IVFSearchParameters {
     size_t nprobe;    ///< number of probes at query time
     size_t max_codes; ///< max nb of codes to visit to do a query
-    IVFSearchParameters() : nprobe(1), max_codes(0) {}
+    int parallel_mode; // default value if -1, and we will use
+                       // this->parallel_mode in this case
+    IVFSearchParameters() : nprobe(1), max_codes(0), parallel_mode(-1) {}
```
cydrain, line 72: set parallel_mode default value to 0

liliu-z: We cannot do that

We have this logic in `xxx_preassigned`:
```
    int preassigned_parallel_mode = 0;
    if (params && params->parallel_mode != -1) {
        preassigned_parallel_mode = params->parallel_mode;
    } else {
        preassigned_parallel_mode = this->parallel_mode;
    }
    int pmode = preassigned_parallel_mode & ~PARALLEL_MODE_NO_HEAP_INIT;
```
the old code is:
```
int pmode = this->parallel_mode & ~PARALLEL_MODE_NO_HEAP_INIT;
```
if default value is 0, assume some other codes pass the param in.
In the old perform: it will use `this->parallel_mode` as parallel_mode
In the new cold if we set default to 0: it will use 0 as parallel_mode. This is unsynced

cydrain (commented):
1. suggest use file name IndexIVF_TS.cpp
2. use name "search_thread_safe", "search_without_codes_thread_safe", and so on
3. move following lines into API GenParams() 
    int parallel_mode = -1;
    if (params->nprobe > 1 && n <= 4) {
        parallel_mode = 1;
    } else {
        parallel_mode = 0;
    }
    size_t max_codes = 0;

Issue #405, opened 2022-08-17
HNSW is not thread safe


liliu-z:
HNSW store the ef in its field and separate the ef setting and search, which introduces race condition.

PR #406, opened 2022-08-17
Make HNSW search thread safe

liliu-z:
 
issue: #405 

HNSW has race condition on `ef_`. It store `ef_` in its field, and separate the `ef_`'s setting and search. This PR is to fix this issue.


liliu-z:
rerun cpu-e2e

PR #415, opened 2022-08-19
Optimize OpenBLAS compiling

matchyc:
Signed-off-by: matchyc <dawnlight.yc@protonmail.com>

Improving OpenBLAS performance by modifying parameters of compilation.
issue: #416 #253 
Why we need this:
First, We have **never used** self-compiled OpenBLAS before.
Then,
1. Original version uses **SINGLE** thread OpenBLAS compiling option which is very slow. We need to enable multi-thread.
2. Eliminate useless modules. i.e. LAPACK or LAPACKE
3. Remove useless compiling options which already set as the default value by OpenBLAS itself.

### DIFF
--- a/cmake/ThirdPartyPackagesKw.cmake
+++ b/cmake/ThirdPartyPackagesKw.cmake
@@ -195,28 +195,26 @@ macro(build_openblas)
 
     set(OPENBLAS_CMAKE_ARGS
             ${EP_COMMON_CMAKE_ARGS}
-            -DVERSION=${OPENBLAS_VERSION}
             -DCMAKE_INSTALL_PREFIX=${KNOWHERE_INSTALL_PREFIX}
-            -DCMAKE_BUILD_TYPE=Release
             -DBUILD_SHARED_LIBS=ON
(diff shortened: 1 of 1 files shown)


cmake/ThirdPartyPackagesKw.cmake
```diff
@@ -195,28 +195,26 @@ macro(build_openblas)
 
     set(OPENBLAS_CMAKE_ARGS
             ${EP_COMMON_CMAKE_ARGS}
-            -DVERSION=${OPENBLAS_VERSION}
             -DCMAKE_INSTALL_PREFIX=${KNOWHERE_INSTALL_PREFIX}
-            -DCMAKE_BUILD_TYPE=Release
             -DBUILD_SHARED_LIBS=ON
-            -DBUILD_STATIC_LIBS=ON
```
Presburger, line 202: static libs may be better for using.

matchyc: OpenBLAS depends on many other libs like omp, math, etc. And we can control linking behaviors when using dynamic libs. Or insist on using a static one?

PR #417, opened 2022-08-19
Support Feder for HNSW and IVFFlat

cydrain:
Issue: #428

### DIFF
--- a/knowhere/feder/HNSW.h
+++ b/knowhere/feder/HNSW.h
@@ -1,0 +1,212 @@
+// Copyright (C) 2019-2020 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under the License
+
+#pragma once
+
+#include <memory>
+#include <tuple>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+#include <nlohmann/json.hpp>
+
+namespace knowhere::feder::hnsw {
+
+///////////////////////////////////////////////////////////////////////////////
+// index view
+struct NodeInfo {
+    int64_t id_;
+    std::vector<int64_t> neighbors_;
+
+    NodeInfo() = default;
+
+    explicit NodeInfo(const int64_t id, std::vector<int64_t>&& neighbors) : id_(id), neighbors_(std::move(neighbors)) {
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(NodeInfo, id_, neighbors_)
+};
+
+class LevelLinkGraph {
+ public:
+    LevelLinkGraph() = default;
+
+    explicit LevelLinkGraph(int64_t level) : level_(level) {
+    }
+
+    int64_t
+    GetLevel() {
+        return level_;
+    }
+
+    const std::vector<NodeInfo>&
+    GetNodes() const {
+        return nodes_;
+    }
+
+    void
+    AddNodeInfo(int64_t id, std::vector<int64_t>&& links) {
+        nodes_.emplace_back(NodeInfo(id, std::move(links)));
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(LevelLinkGraph, level_, nodes_)
+
+ private:
+    int64_t level_;
+    std::vector<NodeInfo> nodes_;
+};
+
+class HNSWMeta {
+ public:
+    HNSWMeta() = default;
+
+    explicit HNSWMeta(int64_t ef_construction,
+             int64_t M,
+             int64_t num_elem,
+             int64_t num_levels,
+             int64_t enter_point_id,
+             int64_t num_overview_levels)
+        : ef_construction_(ef_construction),
+          M_(M),
+          num_elem_(num_elem),
+          num_levels_(num_levels),
+          enter_point_id_(enter_point_id),
+          num_overview_levels_(num_overview_levels) {
+    }
+
+    int64_t
+    GetEfConstruction() {
+        return ef_construction_;
+    }
+
+    int64_t
+    GetM() {
+        return M_;
+    }
+
+    int64_t
+    GetNumElem() {
+        return num_elem_;
+    }
+
+    int64_t
+    GetNumLevel() {
+        return num_levels_;
+    }
+
+    int64_t
+    GetEnterPointId() {
+        return enter_point_id_;
+    }
+
+    int64_t
+    GetNumOverviewLevels() {
+        return num_overview_levels_;
+    }
+
+    const std::vector<LevelLinkGraph>&
+    GetOverviewHierGraph() const {
+        return overview_hier_graph_;
+    }
+
+    void
+    AddLevelLinkGraph(int64_t level) {
+        overview_hier_graph_.push_back(LevelLinkGraph(level));
+    }
+
+    void
+    AddNodeInfo(int64_t level, int64_t id, std::vector<int64_t>&& links) {
+        auto& curr = overview_hier_graph_.back();
+        assert(curr.GetLevel() == level);
+        curr.AddNodeInfo(id, std::move(links));
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(HNSWMeta, ef_construction_, M_, num_elem_, num_levels_, enter_point_id_,
+                                   num_overview_levels_, overview_hier_graph_);
+
+ private:
+    int64_t ef_construction_;
+    int64_t M_;
+
+    int64_t num_elem_;
+    int64_t num_levels_;
+
+    int64_t enter_point_id_;
+    int64_t num_overview_levels_;
+
+    std::vector<LevelLinkGraph> overview_hier_graph_;
+};
+
+///////////////////////////////////////////////////////////////////////////////
+// search view
+class LevelVisitRecord {
+ public:
+    LevelVisitRecord() = default;
+
+    explicit LevelVisitRecord(int64_t level) : level_(level) {
+    }
+
+    int64_t
+    GetLevel() {
+        return level_;
+    }
+
+    const std::vector<std::tuple<int64_t, int64_t, float>>&
+    GetRecords() const {
+        return records_;
+    }
+
+    void
+    AddVisitRecord(int64_t id_from, int64_t id_to, float distance) {
+        records_.emplace_back(std::make_tuple(id_from, id_to, distance));
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(LevelVisitRecord, level_, records_)
+
+ private:
+    int64_t level_;
+    std::vector<std::tuple<int64_t, int64_t, float>> records_;
+};
+
+class HNSWVisitInfo {
+ public:
+    HNSWVisitInfo() = default;
+
+    const std::vector<LevelVisitRecord>&
+    GetInfos() const {
+        return infos_;
+    }
+
+    void
+    AddLevelVisitRecord(int64_t level) {
+        infos_.emplace_back(LevelVisitRecord(level));
+    }
+
+    void
+    AddVisitRecord(int64_t level, int64_t id_from, int64_t id_to, float dist) {
+        auto& curr = infos_.back();
+        assert(curr.GetLevel() == level);
+        curr.AddVisitRecord(id_from, id_to, dist);
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(HNSWVisitInfo, infos_)
+
+ private:
+    std::vector<LevelVisitRecord> infos_;
+};
+
+struct FederResult {
+    HNSWVisitInfo visit_info_;
+    std::unordered_set<int64_t> id_set_;
+};
+using FederResultUniq = std::unique_ptr<FederResult>;
+
+} // namespace knowhere::feder::hnsw
--- a/knowhere/feder/IVFFlat.h
+++ b/knowhere/feder/IVFFlat.h
@@ -1,0 +1,94 @@
+// Copyright (C) 2019-2020 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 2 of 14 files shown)


knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+    std::vector<int64_t> links_;
+
+    NodeInfo() = default;
+
+    NodeInfo(const int64_t id, const std::vector<int64_t>& links) = delete;
+
+    NodeInfo(const int64_t id, std::vector<int64_t>&& links) : id_(id), links_(links) {
+    }
+
+    NodeInfo(const NodeInfo& other) {
```
liliu-z, line 39: This looks like a by-default copy constructor, so does the copy operator. If we want to keep this method of copy, we don't need to specify them.

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+        id_ = other.id_;
+        links_ = other.links_;
+    }
+
+    NodeInfo(NodeInfo&& other) {
```
liliu-z, line 44: A small trick here, move constructor/operator have to be marked as `noexcept`. Since it is not revertable, which means if an exception occur in move, the half-moved things cannot be reverted like a copy.

liliu-z: But as mentioned before, we may not need this since its perform is be-default.

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    NodeInfo
+    operator=(NodeInfo&& other) {
+        id_ = other.id_;
+        links_ = std::move(other.links_);
+        return *this;
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(NodeInfo, id_, links_)
```
liliu-z, line 63: Do we have a conclusion here that does it use a copy or a move?

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+// index view
+struct NodeInfo {
+    int64_t id_;
+    std::vector<int64_t> links_;
+
+    NodeInfo() = default;
+
+    NodeInfo(const int64_t id, const std::vector<int64_t>& links) = delete;
```
liliu-z, line 34: Can I ask why we need to delete this here? If we simply want to just use `NodeInfo()` as constructor, we can add `explicit` before it, so the implicit convert like `NodeInfo{a, b}` will be forbidden.

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+    operator=(NodeInfo&& other) {
+        id_ = other.id_;
+        links_ = std::move(other.links_);
+        return *this;
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(NodeInfo, id_, links_)
+};
+
+struct LevelLinkGraph {
```
liliu-z, line 66: Same questions as before.

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+    operator=(LevelLinkGraph&& other) {
+        level_ = other.level_;
+        nodes_ = std::move(other.nodes_);
+        return *this;
+    }
+
+    void
+    AddNodeInfo(int64_t id, std::vector<int64_t>& links) {
+        nodes_.emplace_back(NodeInfo(id, std::move(links)));
+    }
```
liliu-z, line 102: Struct is preferred to be used as only a [passive carrier](https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes), suggest to use class here

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(LevelLinkGraph, level_, nodes_)
+};
+
+struct HNSWMeta {
```
liliu-z, line 107: Same as before, class is preferred here. and we don't have to crowd them in a same file. Noticed that we have a very detail namespace like `feder::hnsw`, so no harm to create a folder under `feder` called `hnsw` and put all files there. This is just suggestion, depends on implementers preference.

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+
+    int64_t enter_point_id_;
+    int64_t num_overview_levels_;
+
+    std::vector<LevelLinkGraph> overview_hier_graph_;
+
+    void
+    AddLevelLinkGraph(int64_t level) {
+        overview_hier_graph_.emplace_back(LevelLinkGraph(level));
+    }
```
liliu-z, line 122: Do we need the level? From the overview, looks like we only allow use to add `NodeInfo` to the latest level, so why we need use to give us a level id?

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+    LevelLinkGraph
+    operator=(LevelLinkGraph&& other) {
+        level_ = other.level_;
+        nodes_ = std::move(other.nodes_);
+        return *this;
+    }
+
+    void
+    AddNodeInfo(int64_t id, std::vector<int64_t>& links) {
+        nodes_.emplace_back(NodeInfo(id, std::move(links)));
```
liliu-z, line 101: It looks weird to me to move a l-reference. l-reference means the caller will hold the object, but after move, it is empty. If we are sure we want to move, pass r-reference. 

Another trick is that if we do
```
template<typename L>
void AddNodeInfo(int64_t id, L&& links) {
        nodes_.emplace_back(NodeInfo(id, std::forward<typedecl<L>>(links)));
}
```
the performance is if we pass l-reference, we do copy. And if we pass r-reference, we do move.

Anyway, we need to find out what operation we exactly want to do here.

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,203 @@
...
+// index view
+struct NodeInfo {
+    int64_t id_;
+    std::vector<int64_t> links_;
+
+    NodeInfo() = default;
+
+    NodeInfo(const int64_t id, const std::vector<int64_t>& links) = delete;
+
+    NodeInfo(const int64_t id, std::vector<int64_t>&& links) : id_(id), links_(links) {
```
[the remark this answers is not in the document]
cydrain: fixed

knowhere/feder/HNSW.h
cydrain: change it to default

knowhere/feder/HNSW.h
cydrain: fixed

knowhere/feder/HNSW.h
cydrain: "level id" is needed by front end

knowhere/feder/HNSW.h
cydrain: fixed

knowhere/feder/HNSW.h
cydrain: I prefer placing all HNSW related structures and classes in one file

knowhere/feder/HNSW.h
cydrain: removed

knowhere/feder/HNSW.h
cydrain: updated

knowhere/feder/HNSW.h
cydrain: it uses move

knowhere/feder/IVFFlat.h
```diff
@@ -0,0 +1,76 @@
...
+namespace knowhere::feder::ivfflat {
+
+///////////////////////////////////////////////////////////////////////////////
+// index view
+struct ClusterInfo {
+    int64_t id_;
+    std::vector<int64_t> node_ids_;
+    std::vector<float> centroid_vec_;
+
+    ClusterInfo() = default;
```
liliu-z, line 33: Need explicit here

knowhere/feder/IVFFlat.h
```diff
@@ -0,0 +1,76 @@
...
+        std::copy_n(node_id_addr, node_num, node_ids_.data());
+        std::copy_n(centroid_addr, dim, centroid_vec_.data());
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(ClusterInfo, id_, node_ids_, centroid_vec_)
+};
+
+class IVFFlatMeta {
+ public:
+    IVFFlatMeta() = default;
```
liliu-z, line 51: same here

knowhere/index/vector_offset_index/IndexIVF_NM.cpp
```diff
@@ -243,6 +246,46 @@ IVF_NM::QueryByRange(const DatasetPtr& dataset_ptr, const Config& config, const
...
+    int64_t ntotal = ivf_index->ntotal;
+
+    feder::ivfflat::IVFFlatMeta meta(nlist, dim, ntotal);
+    feder::IDSet id_set;
+
+    size_t cnt = 0;
+    for (int32_t i = 0; i < nlist; i++) {
+        // copy from IndexIVF::search_preassigned_without_codes
+        std::unique_ptr<faiss::InvertedLists::ScopedIds> sids;
+        sids.reset(new faiss::InvertedLists::ScopedIds(ivf_index->invlists, i));
```
liliu-z, line 269: Suggest to use `make_unique`

knowhere/feder/IVFFlat.h
cydrain: cpplint reports error if mark default constructor as explicit

liliu-z: Fair, it is somehow redundant.

knowhere/index/vector_offset_index/IndexIVF_NM.cpp
cydrain: updated

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,162 @@
...
+
+#include "knowhere/feder/IDSet.h"
+
+namespace knowhere::feder::hnsw {
+
+///////////////////////////////////////////////////////////////////////////////
+// index view
+struct NodeInfo {
+    int64_t id_;
+    std::vector<int64_t> links_;
```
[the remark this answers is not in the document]
cydrain: updated

knowhere/index/vector_index/IndexHNSW.cpp
```diff
@@ -166,6 +167,12 @@ IndexHNSW::Query(const DatasetPtr& dataset_ptr, const Config& config, const fais
         }
     }
 
+    feder::hnsw::FederResult* feder_result = nullptr;
+    if (CheckKeyInConfig(config, meta::TRACE_VISIT) && GetMetaTraceVisit(config)) {
+        KNOWHERE_THROW_IF_NOT_MSG(rows == 1, "NQ must be 1 when Feder tracing");
```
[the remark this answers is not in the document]
cydrain: no, we only support Feder for single query

knowhere/index/vector_index/IndexHNSW.cpp
```diff
@@ -325,4 +365,41 @@ IndexHNSW::ClearStatistics() {
...
+        visited.insert(curr_id);
+
+        auto data = index_->get_linklist(curr_id, level);
+        auto size = index_->getListCount(data);
+
+        hnswlib::tableint* datal = (hnswlib::tableint*)(data + 1);
+        std::vector<int64_t> links(size);
+        for (int i = 0; i < size; i++) {
+            hnswlib::tableint cand = datal[i];
+            KNOWHERE_THROW_IF_NOT_FMT((cand >= 0 && cand < index_->max_elements_), "Illegal id %d", cand);
```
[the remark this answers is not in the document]
cydrain: this line is copied from hnswlib, I have removed it

knowhere/index/vector_index/IndexHNSW.cpp
```diff
@@ -325,4 +365,41 @@ IndexHNSW::ClearStatistics() {
...
+            links[i] = cand;
+            if (visited.find(cand) == visited.end()) {
+                q.emplace(cand);
+            }
+        }
+        meta.AddNodeInfo(level, curr_id, std::move(links));
+        id_set.Add(curr_id);
+        id_set.Add(links);
```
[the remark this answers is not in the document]
cydrain: good catch, fixed

knowhere/index/vector_index/IndexHNSW.cpp
```diff
@@ -290,6 +306,30 @@ IndexHNSW::QueryByRange(const DatasetPtr& dataset,
     return GenResultDataset(p_id, p_dist, p_lims);
 }
 
+DatasetPtr
+IndexHNSW::GetIndexMeta() {
+    if (!index_) {
+        KNOWHERE_THROW_MSG("index not initialized or trained");
+    }
+
+    const int64_t hnsw_index_overview_levels = 3;
```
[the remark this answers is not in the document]
cydrain: good suggestion

knowhere/feder/HNSW.h
```diff
@@ -0,0 +1,162 @@
...
+        curr.AddVisitRecord(id_from, id_to, dist);
+    }
+
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(HNSWVisitInfo, infos_)
+
+ private:
+    std::vector<LevelVisitRecord> infos_;
+};
+
+struct FederResult {
```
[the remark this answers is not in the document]
cydrain: it's not search result, it records visit info.
FederResult is fine for me.

knowhere/index/vector_index/IndexHNSW.cpp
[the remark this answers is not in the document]
cydrain: updated

knowhere/feder/IDSet.h
```diff
@@ -0,0 +1,47 @@
+// Copyright (C) 2019-2020 Zilliz. All rights reserved.
```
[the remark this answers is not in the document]
cydrain: it's ok, have use std::unordered_set<int64_t> instead

knowhere/index/vector_index/IndexHNSW.cpp
```diff
@@ -231,6 +238,15 @@ IndexHNSW::Query(const DatasetPtr& dataset_ptr, const Config& config, const fais
     // LOG_KNOWHERE_DEBUG_ << "IndexHNSW::Query finished, show statistics:";
     // LOG_KNOWHERE_DEBUG_ << GetStatistics()->ToString();
 
+    // set visit_info json string into result dataset
+    if (feder_result != nullptr) {
+        Config json_visit_info, json_id_set;
+        nlohmann::to_json(json_visit_info, feder_result->visit_info_);
+        nlohmann::to_json(json_id_set, feder_result->id_set_);
+        delete feder_result;
```
liliu-z, line 246: Suggest to use `unique_ptr` here. For `searchKnn` we can pass `const unique_ptr<>&`. This way we can avoid the raw pointer and new/delete

cydrain: updated

PR #432, opened 2022-08-26
Add raw data files and index files check for diskann

cqy123456:
Add some files check in diskann build process:
1. If raw data file not exists, throw exception;
2. If index files exist before building, throw exception.


liliu-z (approved):
/lgtm

Issue #91, opened 2022-03-15
Support RangeSearch


cydrain:
Expose range_search interface for all faiss indices in knowhere:

- [x] faiss idmap
- [x] faiss binary idmap
- [x] faiss ivf
- [x] faiss binary ivf
- [x] ivf_nm

cydrain:
Currently IndexIDMAP and IndexBinaryIDMAP support QueryByDistance, committed from following PR:
https://github.com/milvus-io/milvus/commit/e33d0a797c0cc293a202150755bbd469443c1b6f

PR #433, opened 2022-08-26
Add parameter legacy check API CheckRangeSearch()

cydrain:
Issue: #91

### DIFF
--- a/knowhere/index/vector_index/ConfAdapter.cpp
+++ b/knowhere/index/vector_index/ConfAdapter.cpp
@@ -98,6 +98,14 @@ ConfAdapter::CheckSearch(Config& cfg, const IndexType type, const IndexMode mode
     return true;
 }
 
+bool
+ConfAdapter::CheckRangeSearch(Config& cfg, const IndexType type, const IndexMode mode) {
+    const float DEFAULT_MIN_RADIUS = 0.0;
+    const float DEFAULT_MAX_RADIUS = std::numeric_limits<float>::max();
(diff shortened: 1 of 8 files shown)


knowhere/index/vector_index/ConfAdapter.cpp
```diff
@@ -98,6 +98,14 @@ ConfAdapter::CheckSearch(Config& cfg, const IndexType type, const IndexMode mode
     return true;
 }
 
+bool
+ConfAdapter::CheckRangeSearch(Config& cfg, const IndexType type, const IndexMode mode) {
+    const float DEFAULT_MIN_RADIUS = 0.0;
+    const float DEFAULT_MAX_RADIUS = std::numeric_limits<float>::max();
```
liliu-z, line 104: Why we do this in function constant declaration? I guess we can do it out of the function, and we are trying to use `kSnakeCase` for constant declaration to avoid the conflict with macros

liliu-z: for these kinds of constant declaration, if it only get used here. we can initiate a anonymous namespace for it in the cpp file

liliu-z:
talked offline, this style problem is legacy. For now we just keep it to make the style synced, will change it in 2.0
/lgtm

knowhere/index/vector_index/ConfAdapter.cpp
cydrain: These variables are only used in this API, so we declare here.
Suggest to change this style in knowhere-2.0

Issue #437, opened 2022-08-30
Mixed use of free/delete and new/free in diskann


hhy3:
When built with address sanitizer it'll report this bug.
![20220830-133938](https://user-images.githubusercontent.com/44047980/187358198-31b2ca54-50bf-48a9-add3-18c6364d34e2.jpg)
![20220830-133943](https://user-images.githubusercontent.com/44047980/187358204-1d3a4311-35c0-48ba-8d94-8f2dc826bc30.jpg)

PR #438, opened 2022-08-30
Fix diskann memory problem

hhy3:

issue: #437 

This PR: 
1. Fixed the mixed use of malloc/delete and new/free.
2. Upgrade C++ standard to from C++14 to C++17 in DiskANN.

### DIFF
--- a/knowhere/index/vector_index/IndexDiskANN.cpp
+++ b/knowhere/index/vector_index/IndexDiskANN.cpp
@@ -373,8 +373,8 @@ IndexDiskANN<T>::Query(const DatasetPtr& dataset_ptr, const Config& config, cons
 
     GET_TENSOR_DATA_DIM(dataset_ptr);
     auto query = static_cast<const T*>(p_data);
-    auto p_id = static_cast<int64_t*>(malloc(sizeof(int64_t) * k * rows));
-    auto p_dist = static_cast<float*>(malloc(sizeof(float) * k * rows));
+    auto p_id = new int64_t[k * rows];
+    auto p_dist = new float[k * rows];
 
     std::optional<KnowhereException> ex = std::nullopt;
 #pragma omp parallel for schedule(dynamic, 1)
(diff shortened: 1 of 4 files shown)


liliu-z:
> Signed-off-by: zh Wang [zihao.wang@zilliz.com](mailto:zihao.wang@zilliz.com)
> 
> issue: #437
> 
> This PR fixed the mixed use of malloc/delete and new/free.
> 
> It's better to use std::any instead of boost::any, but it turns out that currently diskann doesn't support c++17. Maybe we should upgrade the standard to c++17.

What do you mean by doesn't support c++17. Knowhere is using C++17 right now.

knowhere/index/vector_index/IndexDiskANN.cpp
```diff
@@ -373,8 +373,8 @@ IndexDiskANN<T>::Query(const DatasetPtr& dataset_ptr, const Config& config, cons
 
     GET_TENSOR_DATA_DIM(dataset_ptr);
     auto query = static_cast<const T*>(p_data);
-    auto p_id = static_cast<int64_t*>(malloc(sizeof(int64_t) * k * rows));
```
matchyc, line 376: Check this response, maybe it's not a good idea?
https://github.com/milvus-io/knowhere/pull/298#discussion_r926881741

hhy3:
> What do you mean by doesn't support c++17. Knowhere is using C++17 right now.

Knowhere is using C++17 but inside thirdparty/DiskANN it only supports C++14 now https://github.com/milvus-io/knowhere/blob/main/thirdparty/DiskANN/CMakeLists.txt#L26 and it can't compile when I use C++17 feature inside thirdparty/DiskANN

knowhere/index/vector_index/IndexDiskANN.cpp
hhy3: In #350 konwhere has replaced malloc/free with new/delete. So it may not be a problem.

PR #442, opened 2022-08-31
Limit the Diskann cached nodes number

cqy123456:

issue: #253 

1.  The number of caches does not exceed 1/3 of the total data size;
2. Remove the code for cache resizing in cache_bfs_levels().


liliu-z:
\lgtm
\approve

PR #454, opened 2022-09-06
Add api to test multithreading

smellthemoon:
Signed-off-by: lixinguo <xinguo.li@zilliz.com>

### DIFF
--- a/unittest/MutiThreadTest.h
+++ b/unittest/MutiThreadTest.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <future>
+#include <vector>
+
+#include "knowhere/common/Dataset.h"
+#include "knowhere/index/VecIndex.h"
+
+namespace knowhere {
+
+namespace mthreadtest {
+std::vector<std::future<DatasetPtr>>&
+GetFeatureQueue() {
+    static std::vector<std::future<DatasetPtr>> qu;
+    return qu;
+}
+
+DatasetPtr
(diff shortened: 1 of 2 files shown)


unittest/MutiThreadTest.h
```diff
@@ -0,0 +1,29 @@
...
+
+namespace mthreadtest {
+std::vector<std::future<DatasetPtr>>&
+GetFeatureQueue() {
+    static std::vector<std::future<DatasetPtr>> qu;
+    return qu;
+}
+
+DatasetPtr
+Async(VecIndexPtr& idx, DatasetPtr& dataset, Config& cfg, faiss::BitsetView bitset) {
```
liliu-z, line 19: I guess we tried to make an API for test team to run code in parallel? If so, suggest to make it as general as possible, which means no any index related thing here. We can try to pass a lambda in.

unittest/MutiThreadTest.h
```diff
@@ -0,0 +1,29 @@
...
+    std::vector<std::future<DatasetPtr>>& qu = GetFeatureQueue();
```
liliu-z, line 20: Can I ask why we need this?

smellthemoon: To store the results, it seems like the test team do the concurrent operations, and the results were obtained.

unittest/MutiThreadTest.h
smellthemoon: Ok, I will have a try.

PR #459, opened 2022-09-08
Add api for multithreading test

smellthemoon:
Signed-off-by: lixinguo <xinguo.li@zilliz.com>
#456

### DIFF
--- a/python/knowhere/__init__.py
+++ b/python/knowhere/__init__.py
@@ -58,6 +58,26 @@ def CreateIndexDiskANN(index_name, index_prefix, metric_type, simd_type="auto"):
             'diskann_f' 'diskann_i8' 'diskann_ui8'."""
     )
 
+def CreateAsyncIndex(index_name, simd_type="auto", *args):
+
+    if simd_type not in ["auto", "avx512", "avx2", "avx", "sse4_2"]:
+        raise ValueError("simd type only support auto avx512 avx2 avx sse4_2")
+
+    if index_name not in ["bin_flat", "bin_ivf_flat", "flat", "ivf_flat", "ivf_pq", "ivf_sq8",
+                          "hnsw", "annoy","gpu_flat", "gpu_ivf_flat",
+                          "gpu_ivf_pq", "gpu_ivf_sq8", "diskann_f", "diskann_i8", "diskann_ui8"]:
+        raise ValueError(
+        """ index name only support
+            'bin_flat', 'bin_ivf_flat', 'flat', 'ivf_flat', 'ivf_pq', 'ivf_sq8',
+            'hnsw', 'annoy', 'gpu_flat', 'gpu_ivf_flat',
+            'gpu_ivf_pq', 'gpu_ivf_sq8', 'diskann_f', 'diskann_i8', 'diskann_ui8'."""
+    )
+    if index_name in ["diskann_f", "diskann_i8", "diskann_ui8"] :
+        index_prefix = args[0]
+        metric_type = args[1]
+        return AsyncIndex(index_name, index_prefix, metric_type) 
+    return AsyncIndex(index_name)
+
 
 class GpuContext:
     def __init__(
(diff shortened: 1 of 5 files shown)


unittest/MultiVectorIndex.h
```diff
@@ -0,0 +1,139 @@
...
+#include "knowhere/index/vector_index/IndexDiskANN.h"
+#include "knowhere/index/vector_index/IndexDiskANNConfig.h"
+#endif
+
+namespace knowhere {
+
+class MultiVectorIndex : public VecIndex {
+ public:
+    MultiVectorIndex(std::string type) {
+        if (type == "annoy") {
```
liliu-z, line 33: use `IndexEnum` instead of raw string

unittest/MultiVectorIndex.h
```diff
@@ -0,0 +1,139 @@
...
+            MultiVectorIndexPtr = std::make_shared<knowhere::IVFPQ>();
+        } else if (type == "ivf_sq8") {
+            MultiVectorIndexPtr = std::make_shared<knowhere::IVFSQ>();
+        } else {
+            KNOWHERE_THROW_FORMAT("Invalid index type %s", std::string(type).c_str());
+        }
+    }
+
+#ifdef KNOWHERE_WITH_DISKANN
+    MultiVectorIndex(std::string type, MetricType metric_type, std::shared_ptr<FileManager> file_manager) {
```
liliu-z, line 55: Did you tried `MultiVectorIndex(std::string type, ...)`?

unittest/MultiVectorIndex.h
```diff
@@ -0,0 +1,139 @@
...
+        return MultiVectorIndexPtr->Dim();
+    }
+
+    int64_t
+    Size() override {
+        return MultiVectorIndexPtr->Size();
+    }
+
+ private:
+    VecIndexPtr MultiVectorIndexPtr;
```
liliu-z, line 137: use `snake_case` here

liliu-z: use `unique_ptr` here

unittest/MultiVectorIndex.h
```diff
@@ -0,0 +1,139 @@
...
+#ifdef KNOWHERE_WITH_DISKANN
+#include "knowhere/index/vector_index/IndexDiskANN.h"
+#include "knowhere/index/vector_index/IndexDiskANNConfig.h"
+#endif
+
+namespace knowhere {
+
+class MultiVectorIndex : public VecIndex {
+ public:
+    MultiVectorIndex(std::string type) {
```
liliu-z, line 32: BTW, did we consider pass a smart pointer to the inner index here?

unittest/MultiVectorIndex.h
smellthemoon: I tried, but it will reduce some cite promblem for reason that, the IndexDiskANN.h was separate from the other indexes when build.

smellthemoon: But I did some changes in the python api , it only has one function to build multivector by using python.

unittest/MultiVectorIndex.h
```diff
@@ -0,0 +1,141 @@
...
+#endif
+
+    std::vector<std::future<DatasetPtr>>&
+    GetFeatureStack() {
+        static std::vector<std::future<DatasetPtr>> stk;
+        return stk;
+    }
+
+    void
+    QueryAsync(const DatasetPtr& dataset, const Config& config, const faiss::BitsetView bitset) {
```
liliu-z, line 77: This is a little bit weird to me:
1. If we want this class to extend `VecIndex`, then we need to use it as a `VecIndex` and don't give extra public API. 
2. If we need to support another group of API, we don't make the inheritance.

In our case, all APIs from `VecIndex` is useless. We can:
1. If swig support function param, we can simple rename this class to something has nothing to do with Index but async like `AsyncRequestEmitter`, and only keep `QueryASync`, `Sync`. `QueryAsync` will accept a function and do async call here.
2. If swig doesn't support that. We can rename it to something like `AsyncIndexProxy`, and remove all `VecIndex` APIs.

PR #471, opened 2022-09-15
Get dim from HNSW index

cydrain:

### DIFF
--- a/knowhere/index/vector_index/IndexHNSW.cpp
+++ b/knowhere/index/vector_index/IndexHNSW.cpp
@@ -252,14 +252,13 @@ IndexHNSW::QueryImpl(int64_t n,
         feder = std::make_unique<feder::hnsw::FederResult>();
     }
 
-    auto dim = GetMetaDim(config);
     size_t ef = GetIndexParamEf(config);
     hnswlib::SearchParam param{ef};
     bool transform = (index_->metric_type_ == 1);  // InnerProduct: 1
 
 #pragma omp parallel for
     for (unsigned int i = 0; i < n; ++i) {
-        auto single_query = xq + i * dim;
+        auto single_query = xq + i * Dim();
         std::priority_queue<std::pair<float, hnswlib::labeltype>> rst;
         auto dummy_stat = hnswlib::StatisticsInfo();
         rst = index_->searchKnn(single_query, k, bitset, dummy_stat, &param, feder);
@@ -298,7 +297,6 @@ IndexHNSW::QueryByRangeImpl(int64_t n,
         feder = std::make_unique<feder::hnsw::FederResult>();
     }
 
-    auto dim = GetMetaDim(config);
     auto range_k = GetIndexParamHNSWK(config);
     size_t ef = GetIndexParamEf(config);
     hnswlib::SearchParam param{ef};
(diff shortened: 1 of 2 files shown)


knowhere/index/vector_index/IndexHNSW.cpp
```diff
@@ -252,7 +252,7 @@ IndexHNSW::QueryImpl(int64_t n,
         feder = std::make_unique<feder::hnsw::FederResult>();
     }
 
-    auto dim = GetMetaDim(config);
+    auto dim = index_->space_->get_data_size() / sizeof(float);
```
liliu-z, line 255: Why we don't pass dim in from `Query()` ?

liliu-z (commented):
1. Can I ask why we make `QueryImp` and `RangeQueryImp` protected instead of private?
2. I noticed that we remove statistics related code. Can I ask what is the status of this features? If we need to remove them, suggest to remove them from all places(Knowhere & Milvus). If we only do this in HNSW, it brings consistency and confusion.

cydrain:
> 1. Can I ask why we make `QueryImp` and `RangeQueryImp` protected instead of private?
QueryImpl() exists for many index types, such as IVF serial. For them, QueryImpl() is protected, so HNSW also mark it as protected.

> 2. I noticed that we remove statistics related code. Can I ask what is the status of this features? If we need to remove them, suggest to remove them from all places(Knowhere & Milvus). If we only do this in HNSW, it brings consistency and confusion.
Statistics related code is disabled since faiss upgraded from 1.6.3 to 1.7.2, I prefer to clean all of them from knowhere using another PR.

knowhere/index/vector_index/IndexHNSW.cpp
cydrain: It's better to get dim from index

liliu-z:
> > 1. Can I ask why we make `QueryImp` and `RangeQueryImp` protected instead of private?
> >    QueryImpl() exists for many index types, such as IVF serial. For them, QueryImpl() is protected, so HNSW also mark it as protected.
> 
> > 2. I noticed that we remove statistics related code. Can I ask what is the status of this features? If we need to remove them, suggest to remove them from all places(Knowhere & Milvus). If we only do this in HNSW, it brings consistency and confusion.
> >    Statistics related code is disabled since faiss upgraded from 1.6.3 to 1.7.2, I prefer to clean all of them from knowhere using another PR.

For the Statistics change, I am good to keep it what it is right now. We can set another PR to clean them all to make knowhere consistent on this.

For setting the QueryImp protected, this is not a good reason:
1. Protected in other indexes is due to inheritance purpose (Although it is not a good design, but implement is reasonable). And we don't have any inheritance here.
2. We need to mark all functions that will not get used outside this class as private.

knowhere/index/vector_index/IndexHNSW.cpp
liliu-z: Then we have three place to get dim:
1. From `DataSet` passed in.
2. From `Dim()` function in `IndexHNSW`
3. From `index_->space_->get_data_size() / sizeof(float)`

this is too confusing, can we ensure they are consistent and which one should we pick?

PR #472, opened 2022-09-15
Optimize lapack library

Presburger:
Signed-off-by: Yusheng.Ma <Yusheng.Ma@zilliz.com>


ci/docker/ubuntu/scripts/deps/build.sh
```diff
@@ -1,19 +1,39 @@
 CMAKE_VERSION="v3.23"
```
jennyli-z, line 1: This can be removed also?

PR #493, opened 2022-10-08
Re-implement HNSW range search

cydrain:

### DIFF
--- a/knowhere/index/vector_index/IndexHNSW.cpp
+++ b/knowhere/index/vector_index/IndexHNSW.cpp
@@ -297,7 +297,6 @@ IndexHNSW::QueryByRangeImpl(int64_t n,
         feder = std::make_unique<feder::hnsw::FederResult>();
     }
 
-    auto range_k = GetIndexParamHNSWK(config);
     size_t ef = GetIndexParamEf(config);
     hnswlib::SearchParam param{ef};
     bool is_IP = (index_->metric_type_ == 1);  // InnerProduct: 1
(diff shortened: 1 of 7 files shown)


PwzXxm:
This modified algorithm obtains range search results by traversing like KNN and the length of results is equal to `ef`. Since there is no quantization and loss in data representation, it will work just fine. It also removes the parameters introduced in the old range search algorithm, which is great. However, a few considerations might be worth mention

   1. Using `priority_queue` to store the results might have performance degradations when the number of results is large as each insert takes logarithmic time. The amount of degradation depends on the number of results, the segment's size, and the distribution of the dataset.
   2. If the connectivity of the graph is low, some results might be missing since there is only one point that comes down to the base layer.

PwzXxm:
Otherwise, /lgtm till tests fixed.

cydrain:
> This modified algorithm obtains range search results by traversing like KNN and the length of results is equal to `ef`. Since there is no quantization and loss in data representation, it will work just fine. It also removes the parameters introduced in the old range search algorithm, which is great. However, a few considerations might be worth mention
> 
> 1. Using `priority_queue` to store the results might have performance degradations when the number of results is large as each insert takes logarithmic time. The amount of degradation depends on the number of results, the segment's size, and the distribution of the dataset.
The new API searchRange() is modified based on API searchKnn(), only 2 lines changed.
Using 'priority_queue' can make the result sorted, it will save the sorting time outside.

> 2. If the connectivity of the graph is low, some results might be missing since there is only one point that comes down to the base layer.
This behavior will be similar as general search. If the connectivity of the graph is low, search will also be not able to get topk results.

cydrain:
Test with sift1m dataset

old
hnsw_range_old

new
hnsw_range_new

The new range search algorithm has more flat performance, and we can improve the recall rate by increase the parameter 'ef'.

The test baseline need update.

PR #496, opened 2022-10-09
Support metric IP for diskann range search

cqy123456:
Diskann uses L2 metirc to caculate IP distance, we can convert the result to the origin ip distance by this formula：
image
issue: https://github.com/milvus-io/knowhere/issues/253

### DIFF
--- a/thirdparty/DiskANN/src/pq_flash_index.cpp
+++ b/thirdparty/DiskANN/src/pq_flash_index.cpp
@@ -1199,8 +1199,8 @@ namespace diskann {
       if (distances != nullptr) {
         distances[i] = full_retset[i].distance;
         if (metric == diskann::Metric::INNER_PRODUCT) {
-          // flip the sign to convert min to max
-          distances[i] = (-distances[i]);
+          // convert l2 distance to ip distance
+          distances[i] = 1.0 - distances[i] / 2.0;
           // rescale to revert back to original norms (cancelling the effect of
           // base and query pre-processing)
           if (max_base_norm != 0)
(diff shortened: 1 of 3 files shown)


unittest/test_diskann.cpp
```diff
@@ -384,15 +403,17 @@ float* DiskANNTest::global_large_dim_query_data_ = nullptr;
...
 GroundTruthPtr DiskANNTest::large_dim_l2_ground_truth_ = nullptr;
 GroundTruthPtr DiskANNTest::large_dim_l2_range_search_ground_truth_ = nullptr;
+GroundTruthPtr DiskANNTest::large_dim_ip_range_search_ground_truth_ = nullptr;
 
 INSTANTIATE_TEST_CASE_P(DiskANNParameters, DiskANNTest,
-                        Values(std::make_tuple(knowhere::metric::L2, true /* large dim */),
-                               std::make_tuple(knowhere::metric::L2, false /* large dim */),
-                               std::make_tuple(knowhere::metric::IP, true /* large dim */),
-                               std::make_tuple(knowhere::metric::IP, false /* large dim */)));
+                        Values(std::make_tuple(knowhere::metric::L2, true /* high-dimension */),
```
liliu-z, line 413: Good catch!

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -1246,10 +1246,18 @@ namespace diskann {
...
+            res_count = i;
+            break;
+          }
+        } else {
+          if (distances[i] > (float) range || indices[i] == -1) {
+            res_count = i;
+            break;
+          }
+        }
+        if (i == l_search - 1)
```
liliu-z, line 1260: Two problems:
1. this embedded `if` logic's readability is too low.
2. we need to check `indces[i] == -1` first. Since `distances[i] == -1` doesn't means `indices[i] != -1`.

Suggest to make it 
```
      if (indices[i] == -1) {
        res_count = i;
        break;
      }
      bool out_of_range = metric == diskann::Metric::INNER_PRODUCT
                              ? distances[i] < (float) range
                              : distances[i] > (float) range;
      if (out_of_range) {
        res_count = i;
        break;
      }
      if (i == l_search - 1) {
        res_count = l_search;
      }
```

unittest/test_diskann.cpp
```diff
@@ -827,3 +839,42 @@ TEST_P(DiskANNTest, generate_cache_list_test) {
     knowhere::DiskANNPrepareConfig::Set(cfg, prep_conf_to_test);
     EXPECT_THROW(diskann->Prepare(cfg), knowhere::KnowhereException);
 }
+
+TEST_P(DiskANNTest, result_distance_test) {
```
liliu-z, line 843: We can add distance check to all other UT instead of adding a brand new one.

we can make the `dis_truth` as a global variables like `ground_truth`, and calculate it at the beginning time.

liliu-z:
> @cqy123456 Please associate the related issue to the body of your Pull Request. (eg. “issue: #”)

PR #502, opened 2022-10-14
Remove assemble and disassemble util func

xige-16:
/kind improvement

knowhere does not need to perceive how to slice and merge index data, how to slice index should be handled by upper-level business logic

Signed-off-by: xige-16 <xi.ge@zilliz.com>


cydrain:
following definition should be removed
constexpr const char* SLICE_SIZE = "SLICE_SIZE";

PR #509, opened 2022-10-25
Remove HNSW range search parameter range_k, use ef instead

cydrain:

before:
hnsw_range_search_old_1025

with this change:
hnsw_range_search_new_1025


PwzXxm:
As we discussed in the meeting. Waiting for e2e tests pass.

cydrain:
Since HNSW range search algorithm updated, the recall baseline need also be updated.
Will ask @yanliang567 to help on this.

PR #517, opened 2022-11-01
Make the Query thread pool global

liliu-z:

This is PR is to make the thread pool global and shared by different segment in one query node. 

We did a test on 2C8G container with 6M SIFT data, which is 8 segments. The improvement of serial/parallel execution is about ~10-15%

issue: #513

### DIFF
--- a/knowhere/common/ThreadPoolHolder.cpp
+++ b/knowhere/common/ThreadPoolHolder.cpp
@@ -1,0 +1,47 @@
+// Copyright (C) 2019-2020 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under the License
+
+#include "knowhere/common/ThreadPoolHolder.h"
+
+#include "knowhere/common/Log.h"
+
+namespace knowhere {
+
+void
+ThreadPoolHolder::Init(uint32_t num_threads) {
+    if (!pool_) {
+        std::lock_guard<std::mutex> lock(mutex_);
+        if (!pool_) {
+            pool_ = std::make_shared<ctpl::thread_pool>(num_threads);
+        }
+    }
+    LOG_KNOWHERE_WARNING_ << "ThreadPoolHolder has already been initilaized with threads num: " << pool_->size();
+}
+
+std::shared_ptr<ctpl::thread_pool>
+ThreadPoolHolder::GetThreadPool() {
+    if (!pool_) {
+        std::lock_guard<std::mutex> lock(mutex_);
+        if (!pool_) {
+            LOG_KNOWHERE_WARNING_ << "ThreadPoolHolder is not inialized, init it with threads num: "
+                                  << std::thread::hardware_concurrency();
+            pool_ = std::make_shared<ctpl::thread_pool>(std::thread::hardware_concurrency());
+        }
+    }
+    return pool_;
+}
+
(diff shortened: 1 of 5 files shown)


knowhere/common/ThreadPool.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "ctpl/ctpl-std.h"
+
+namespace knowhere {
+
+static std::shared_ptr<ctpl::thread_pool> pool_;
+static std::mutex mutex;
+
+void
+ThreadPool::Init(uint32_t num_threads) {
```
cydrain, line 25: Since the 2 APIs in this file are very short, suggest to write them to ThreadPool.h

knowhere/common/ThreadPool.cpp
```diff
@@ -0,0 +1,40 @@
...
+    std::lock_guard<std::mutex> lock(mutex);
+    if (!pool_) {
+        pool_ = std::make_shared<ctpl::thread_pool>(num_threads);
+    }
+}
+
+std::shared_ptr<ctpl::thread_pool>
```
cydrain, line 32: suggest to change this API in this way, "mutex" is not necessary.
```
std::shared_ptr<ctpl::thread_pool>
ThreadPool::GetInstance() {
    std::shared_ptr<ctpl::thread_pool> pool_ = std::make_shared<ctpl::thread_pool>(std::thread::hardware_concurrency());
    return pool_;
}
```

liliu-z: > suggest to change this API in this way, "mutex" is not necessary.
> 
> ```
> std::shared_ptr<ctpl::thread_pool>
> ThreadPool::GetInstance() {
>     std::shared_ptr<ctpl::thread_pool> pool_ = std::make_shared<ctpl::thread_pool>(std::thread::hardware_concurrency());
>     return pool_;
> }
> ```

First the `pool_` has to be static, so we need to make it static here.
Then, we need to expose an API allowing user to customize the thread_num, so simply change it to `Meyers Singleton` is not suitable here.

knowhere/common/ThreadPoolHolder.cpp
```diff
@@ -0,0 +1,55 @@
...
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (!pool_) {
+        pool_ = std::make_shared<ctpl::thread_pool>(num_threads);
+    } else {
+        LOG_KNOWHERE_WARNING_ << "ThreadPoolHolder has already been initilaized with threads num: "
+                              << pool_->size();
+    }
+}
+
+std::shared_ptr<ctpl::thread_pool>
```
cydrain, line 35: std::shared_ptr<ctpl::thread_pool>
ThreadPoolHolder::GetThreadPool() {
    if (!pool_) {
        std::lock_guard<std::mutex> lock(mutex_);
        if (!pool_) {
            LOG_KNOWHERE_WARNING_ << "ThreadPoolHolder is not inialized, init it with threads num: "
                                  << std::thread::hardware_concurrency();
            pool_ = std::make_shared<ctpl::thread_pool>(std::thread::hardware_concurrency());
        }
    }
    return pool_;
}

knowhere/common/ThreadPoolHolder.cpp
```diff
@@ -0,0 +1,55 @@
...
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under the License
+
+#include "knowhere/common/ThreadPoolHolder.h"
+
+#include "knowhere/common/Log.h"
+
+namespace knowhere {
+
+void
```
cydrain, line 18:
```
void
ThreadPoolHolder::Init(uint32_t num_threads) {
    if (!pool_) {
        std::lock_guard<std::mutex> lock(mutex_);
        if (!pool_) {
            pool_ = std::make_shared<ctpl::thread_pool>(num_threads);
        }
    }
    LOG_KNOWHERE_WARNING_ << "ThreadPoolHolder has already been initilaized with threads num: "
                                  << pool_->size();
}
```

cqy123456:
lgtm

PR #527, opened 2022-11-04
Remove stats in hnsw

liliu-z:

`stats` is not thread safe in Knowhere. Delete it in case of crash.


Presburger:
\lgtm

PR #552, opened 2022-11-21
Optimize hnsw search

hhy3:

issue: #547 

This PR replaces `visited_list_pool` with bitset, which uses less memory footprint, and thus easier to optimize memory prefetch logic. This optimization will gain near 10% performance improvement.

visited


liliu-z (approved):
/lgtm

PR #553, opened 2022-11-21
Optimize hnsw search algo

hhy3:

issue: #547 

This PR replaces `priority_queue` in hnsw search with linear neighbors pool. This'll gain approximately 5% performance improvement when the recall is around 90%. When `ef` becomes larger i.e. when the recall approaches 100%, the difference is diminishing as expected, as the time complexity begins to dominate.

pool


liliu-z (approved):
/lgtm

PR #559, opened 2022-11-22
Add visited list pool in hnsw

hhy3:

issue: #547 

This PR replaces the reusable visited pool which is of poor performance in hnsw with bitset pool


liliu-z (approved):
/lgtm

PR #569, opened 2022-11-29
Add API GenResultDataSet()

cydrain:

### DIFF
--- a/include/knowhere/dataset.h
+++ b/include/knowhere/dataset.h
@@ -211,6 +211,7 @@ class DataSet {
 };
 using DataSetPtr = std::shared_ptr<DataSet>;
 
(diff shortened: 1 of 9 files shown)


include/knowhere/dataset.h
```diff
@@ -221,5 +221,40 @@ GenDataSet(const int64_t nb, const int64_t dim, const void* xb) {
     return ret_ds;
 }
```
Presburger, line 223:
```

#ifdef NOT_COMPILE_FOR_SWIG
---your code----
#endif 
```

then add 
```
add_definitions(-DNOT_COMPILE_FOR_SWIG)
```
into CMakeLists.txt

PR #579, opened 2022-12-02
Add DataSet2Array for range search

cydrain:

### DIFF
--- a/python/knowhere/knowhere.i
+++ b/python/knowhere/knowhere.i
@@ -195,4 +195,22 @@ DataSet2Array(DataSetPtr result, float* dis, int nq_1, int k_1, int* ids, int nq
     }
 }
 
+void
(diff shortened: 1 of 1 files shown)


python/knowhere/knowhere.i
```diff
@@ -195,4 +195,22 @@ DataSet2Array(DataSetPtr result, float* dis, int nq_1, int k_1, int* ids, int nq
     }
 }
 
+void
+DataSet2Array(DataSetPtr result, int* lims, int s_1, float* dis, int s_2, int* ids, int s_3) {
```
Presburger, line 199: need two axis,not ones
DataSet2Array(DataSetPtr result, int* lims, int nq_0,int k_0, float* dis, int nq_1, int k_1, int* ids, int nq_2, int k_2)

add
%apply (int *INPLACE_ARRAY2, int DIM1, int DIM2){(int *lims,int nq_0,int k_0)}
after line 63

PR #616, opened 2022-12-30
Add knowhere::status, bitset, binaryset in python api

cqy123456:
Signed-off-by: cqy123456 <qianya.cheng@zilliz.com>

### DIFF
--- a/python/knowhere/__init__.py
+++ b/python/knowhere/__init__.py
@@ -1,9 +1,14 @@
 from . import swigknowhere
+from .swigknowhere import Status
+from .swigknowhere import GetBinarySet, GetNullDataSet, GetNullBitSetView
 import numpy as np
-
 
 def CreateIndex(name):
     return swigknowhere.IndexWrap(name)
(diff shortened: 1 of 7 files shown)


python/knowhere/__init__.py
```diff
@@ -1,11 +1,24 @@
 from . import swigknowhere
 import numpy as np
 
+Status = swigknowhere.Status
```
Presburger, line 4: from swigknowhere import Status, maybe better?

python/knowhere/__init__.py
```diff
@@ -1,11 +1,24 @@
...
 
 def CreateIndex(name):
     return swigknowhere.IndexWrap(name)
 
 
+def CreateBitSet(bits_num):
+    return swigknowhere.Bitset(bits_num)
+
+
+def GetEmptyBitsetView():
```
Presburger, line 14: from swigknowhere import GetEmptyBitsetView

PR #693, opened 2023-02-15
Fix GPU index build error

cydrain:


matrixji:
Do we have CI jobs to compile with CUDA?

xiaofan-luan:
We will be working on adding one~
GPU support will be the next priority

PR #712, opened 2023-03-02
Provide support for RAFT-based indexes

### LINKED ISSUE #715: Provide support for RAFT Indexes
[RAFT](https://github.com/rapidsai/raft) offers optimized implementations of several nearest-neighbor algorithms on GPU. To provide optimal GPU performance and take advantage of ongoing work in RAFT, it would be useful to have RAFT indexes available as another alternative within Knowhere.

wphicks:
This is still a work-in-progress. Complete PR description will be added once it is ready for review. Resolve #715.

issue: #715


xiaofan-luan:
@wphicks Welcome !

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -0,0 +1,385 @@
...
+                                        data,
+                                        data_gpu.size() * sizeof(float),
+                                        cudaMemcpyDefault,
+                                        stream.value()
+                                        )
+                                    );
+                            if constexpr (std::is_same_v<detail::raft_ivf_flat_index, T>) {
+                                auto build_params = raft::neighbors::ivf_flat::index_params{};
+                                build_params.metric = metric.value();
+                                build_params.n_lists = ivf_raft_cfg.nlist;
```
cjnolet, line 134: I'm assuming there's going to be a parameter sweep in order to tune the settings for computing perf at different recall levels? We should expose all of the params so we can sweep through as much of them as possible- there aren't as many for ivf flat as ivf pq, but it will probably be helpful to expose the number of coarse quantizer iterations and amount of the dataset to use for training the coarse quantizer here.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -0,0 +1,385 @@
...
+                                            build_params,
+                                            data_gpu.data(),
+                                            rows,
+                                            dim
+                                     );
+                            } else if constexpr (std::is_same_v<detail::raft_ivf_pq_index, T>) {
+                                auto build_params = raft::neighbors::ivf_pq::index_params{};
+                                build_params.metric = metric.value();
+                                build_params.n_lists = ivf_raft_cfg.nlist;
+                                build_params.pq_bits = ivf_raft_cfg.nbits;
```
cjnolet, line 148: Should probably expose pq_dim here as well, in addition to the coarse quantizer iterations. If it's possible to do in Knowhere, we should also consider smaller precision sizes for the training/query datasets as well.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -0,0 +1,385 @@
...
+                                    *gpu_index_,
+                                    data_gpu.data(),
+                                    rows,
+                                    ivf_raft_cfg.k,
+                                    ids_gpu.data(),
+                                    dis_gpu.data()
+                                    );
+                        } else if constexpr (std::is_same_v<detail::raft_ivf_pq_index, T>) {
+                            auto search_params = raft::neighbors::ivf_pq::search_params{};
+                            search_params.n_probes = ivf_raft_cfg.nprobe;
```
cjnolet, line 275: The precision of the lookup table can have quite an effect on lookup times. We should probably expose that here.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -0,0 +1,385 @@
...
+                            raft::neighbors::ivf_pq::search<float, std::int64_t>(
```
cjnolet, line 276: Just a heads up, the pointer-based APIs are going to be going away very soon. I'd consider using the mdspan API for IVF Flat since it's available now. We have a PR to add the mdspan API for IVF PQ going in soon. We might be removing the pointer APIs as soon as 23.04.

cjnolet (commented):
Integration looks pretty clean so far! Leaving some feedback as we will probably want to allow sweeping through most, of not all, of the available parameters to get accurate benchmarks.

src/common/raft_metric.h
```diff
@@ -0,0 +1,48 @@
...
+#include "knowhere/expected.h"
+
+namespace knowhere {
+
+inline expected<raft::distance::DistanceType, Status>
+Str2RaftMetricType(std::string metric) {
+    static const std::unordered_map<std::string, raft::distance::DistanceType> metric_map = {
+        {metric::L2, raft::distance::DistanceType::L2Expanded},
+        {metric::IP, raft::distance::DistanceType::InnerProduct},
+        {metric::HAMMING, raft::distance::DistanceType::HammingUnexpanded},
```
cjnolet, line 35: Not sure if it makes a difference here or if Knowhere will just silently skip in the case of an exception but I don't think we support hamming or jaccard on the IVF APIs yet. 

Both of these can be formulated post processing on a dot product, though, so we might be able to support them in a future release.

matrixji:
I see [raft support for faiss](https://github.com/facebookresearch/faiss/pull/2707), right now Knowhere reuses some faiss's code for some GPU Index. Does it possible to reuse the faiss's impl, if Faiss would support RAFT?

liliu-z:
> I see [raft support for faiss](https://github.com/facebookresearch/faiss/pull/2707), right now Knowhere reuses some faiss's code for some GPU Index. Does it possible to reuse the faiss's impl, if Faiss would support RAFT?

We will definitely do that. Since it takes time for Faiss' integration to be done, we can first support RAFT to fillfull users' needs.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -0,0 +1,385 @@
...
+    };
+template <>
+    struct KnowhereConfigType<detail::raft_ivf_pq_index> {
+        typedef RaftIvfPqConfig Type;
+    };
+
+template <typename T>
+    class RaftIvfIndexNode : public IndexNode {
+        public:
+            RaftIvfIndexNode(const Object& object) : devs_{}, res_{std::make_unique<raft::device_resources>()}, gpu_index_{} {}
```
Presburger, line 77: maybe we need a device memory pool.

src/index/ivf_raft/ivf_raft.cuh
wphicks: Agreed! Same thoughts on above in terms of possibly doing this in a follow-up or getting help from other contributors.

wphicks:
> I see [raft support for faiss](https://github.com/facebookresearch/faiss/pull/2707), right now Knowhere reuses some faiss's code for some GPU Index. Does it possible to reuse the faiss's impl, if Faiss would support RAFT?

Yes, as @liliu-z mentioned, integrating RAFT is mostly to make it available within Milvus sooner. It may also be an integration path that will be useful to support longer term, since ANN improvements are a very active area of research and development in RAFT. New innovations there may take a bit of time to be supported through FAISS, so having RAFT support will allow Milvus to more quickly take advantage of the latest and greatest we can come up with.

src/index/ivf_raft/ivf_raft.cuh
wphicks: This is fairly trivial to add thanks to RMM, but we may need some further discussion about how exactly it should behave with respect to the rest of Knowhere. For instance, it is easy to set up a pool that grows to accommodate however much memory is required, but do we need to set up a mechanism to release that memory if other index types (which do not use RMM) are used in combination with an IVF RAFT index? Alternatively, do we want to have a fix-sized memory pool, and if so how should users pass the desired size?

I'm happy to add the simplest version of a memory pool right away to this PR, but if there are more complex considerations, we might want to look at them in a follow-up.

src/index/ivf_raft/ivf_raft.cuh
cjnolet: As long as this first implementation is not assumed to be "benchmarking" complete (e.g. any benchmarking will not be assumed valid) then I think that's fine. Otherwise, at the very least, the additional ivf_pq parameters are going to be needed to achiece competitive performance.

wphicks: Yes, I think you're right. Given that we're going to want to move fairly quickly toward benchmarking this, maybe we'd better expose that sooner rather than later. I'll have a chat with other folks and see how best to pass through the necessary parameters.

src/index/ivf_raft/ivf_raft.cuh
cjnolet: BTW- the explicit rmm memory resource argument for searching is going away soon. However, an rmm memory pool can still be set as the `default` on the current process. The `raft::device_resources` (which is used by all of RAFT's APIs) also contains a `workspace_resource` which is an RMM resource that can be used to override the default RMM resource (for example, for setting a pool memory resource only for temporary device allocations).

Presburger:
> > I see [raft support for faiss](https://github.com/facebookresearch/faiss/pull/2707), right now Knowhere reuses some faiss's code for some GPU Index. Does it possible to reuse the faiss's impl, if Faiss would support RAFT?
> 
> Yes, as @liliu-z mentioned, integrating RAFT is mostly to make it available within Milvus sooner. It may also be an integration path that will be useful to support longer term, since ANN improvements are a very active area of research and development in RAFT. New innovations there may take a bit of time to be supported through FAISS, so having RAFT support will allow Milvus to more quickly take advantage of the latest and greatest we can come up with.

raft allows knowhere to rely on fewer libs, faiss lacks modularity in design, so I prefer raft.

src/index/ivf_raft/ivf_raft.cuh
wphicks: I've just added basic usage of RMM's pool resource, and we can refine from there.

wphicks:
I've just updated with support for a pool allocator and all of RAFT's IVF parameters.

Presburger (approved):

Presburger:
@wphicks hi, if you can merge the commit into single one and signed the commits,  then we will merge the pr.

CMakeLists.txt
```diff
@@ -97,6 +121,11 @@ if(NOT USE_CUDA)
   list(REMOVE_ITEM KNOWHERE_SRCS ${KNOWHERE_GPU_SRCS})
 endif()
 
+if(NOT WITH_RAFT)
+  knowhere_file_glob(GLOB_RECURSE KNOWHERE_RAFT_SRCS src/index/index_raft/*.cc src/index/index_raft/*.cu)
```
matrixji, line 125: seems here `index_raft` should be `ivf_raft`

PR #714, opened 2023-03-02
Let config accept both int and float as float type

cydrain:


include/knowhere/config.h
```diff
@@ -299,7 +299,7 @@ class Config {
                     *ptr->val = ptr->default_val.value();
                     continue;
                 }
-                if (!json[it.first].is_number_float()) {
+                if (!json[it.first].is_number()) {
```
PwzXxm, line 302: The arithmetic overflow check below becomes invalid, and all float values in the config can be float/int, is this acceptable? @liliu-z

liliu-z: Looks good to me. cc @Presburger as the original author.

Presburger: Looks good.

Presburger (approved):

Presburger:
\lgtm

PR #716, opened 2023-03-03
Use shared_ptr for index_ in index node

cydrain:


src/index/ivf/ivf.cc
```diff
@@ -100,9 +100,7 @@ class IvfIndexNode : public IndexNode {
             auto nlist = index_->nlist;
             auto code_size = index_->code_size;
             return (nb * code_size + nb * sizeof(int64_t) + nlist * code_size);
-        }
-
-        if constexpr (std::is_same<T, faiss::IndexIVFPQ>::value) {
+        } else if constexpr (std::is_same<T, faiss::IndexIVFPQ>::value) {
```
Presburger, line 103: not need else if constexpr, this if condition selected after code compiled.
line 45 static_assert is limited the template type.

src/index/ivf/ivf.cc
```diff
@@ -138,30 +134,18 @@ class IvfIndexNode : public IndexNode {
...
-            delete index_;
-        }
-        if (qzr_) {
-            delete qzr_;
-        }
-    };
 
  private:
-    T* index_;
+    std::shared_ptr<T> index_;
```
Presburger, line 147: use T * ptr cause memory leak?

src/index/ivf/ivf.cc
```diff
@@ -219,77 +203,51 @@ IvfIndexNode<T>::Train(const DataSet& dataset, const Config& cfg) {
...
-        } else {
-            index->train(rows, (const float*)data);
-        }
-    } catch (std::exception& e) {
-        delete qzr;
-        delete index;
-        LOG_KNOWHERE_WARNING_ << "faiss inner error, " << e.what();
-        return Status::faiss_inner_error;
-    }
-    if (this->index_) {
```
Presburger, line 283: if train failed, break the old index, maybe not a good choice.

src/index/ivf/ivf.cc
cydrain: There are many places we need check "index_ != nullptr" then do "delete index_", including exception handling and gpu index / cpu index copying, 
if use shared_ptr, we can make coding easier and leak free.

src/index/ivf/ivf.cc
cydrain: ok, roll it back

src/index/ivf/ivf.cc
cydrain: if train fail, "qzr" is deleted manually, "index_" is deleted by system

src/index/ivf/ivf.cc
```diff
@@ -219,77 +209,54 @@ IvfIndexNode<T>::Train(const DataSet& dataset, const Config& cfg) {
...
-                delete qzr;
-            }
-            if (index) {
-                delete index;
-            }
-            LOG_KNOWHERE_WARNING_ << "memory malloc error.";
-            return Status::malloc_error;
+            index_ = std::make_shared<faiss::IndexBinaryIVF>(qzr, dim, nlist, metric.value());
+            index_->own_fields = true;
+            index_->train(rows, (const uint8_t*)data);
```
Presburger, line 250: if this place index_ make_shared failed, or train failed, the old index hold on  var index_ maybe breaked.

cydrain: one segment will only be trained once, "index_" must be nullptr before Train

Presburger (approved):

Presburger:
\lgtm

PR #725, opened 2023-03-09
Set config check log to debug level

### LINKED ISSUE #726: Set config check log to debug level
All the config check log should be debug level log, especially during search which will affect performance.

chasingegg:
Fix #726


chasingegg:
seems still need a approve...

Issue #727, opened 2023-03-09
Use Conan to manage dependency


jiaoew1991:
Milvus has already been using Conan to manage C++ dependencies, and Knowhere can also try to use it.
There are several benefits:
1. Adding dependencies is easier.
2. There will be no conflicts with Milvus' dependencies.
3. Knowhere can be published to JFrog as a package, saving compilation time when compiling Milvus.
4. OpenBLAS can use the package provided by Conan, without the need for maintaining an OpenBLAS image in Milvus.

PR #728, opened 2023-03-09
Use conan to manage depedency

jiaoew1991:
issue: https://github.com/milvus-io/knowhere/issues/727


conanfile.py
```diff
@@ -0,0 +1,184 @@
...
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def requirements(self):
+        self.requires("boost/1.81.0")
+        self.requires("glog/0.6.0")
+        self.requires("nlohmann_json/3.11.2")
+        self.requires("llvm-openmp/12.0.1")
+        if self.options.with_ut:
+            self.requires("gtest/1.8.1")
```
matrixji, line 81: seems previous we use gtest 1.12, AIK gtest<=1.10 may have some compilation problems for the new compiler.

matrixji:
A general discussion about how to use knowhere from the developer's view, recently I'm thinking of this for `milvus-sdk-cpp` also.

As knowhere is a library, it's not fully like milvus, so developers may use knowhere as a component and integrate it into their own project, for milvus currently most developers will only use the final binary.
So as a developer if they could fetch knowhere from conan.io, that's fine, it's pretty good. But someone may want to keep the original way(just use cmake), so if possible, IMO, we'd better provide conan as an option, but not a mandatory thing.

Just like grpc, when I go through [grpc-cpp's README](https://github.com/grpc/grpc/tree/master/src/cpp#readme) it tells me grpc could be imported by Bazal,  Cmake, vcpkg and also conan(which not documented in README)

CMakeLists.txt
```diff
@@ -75,10 +75,24 @@ include_directories(thirdparty/bitset)
...
+  message(STATUS "nlohmann_json not found, fetching from github")
+  Include(FetchContent)
+  FetchContent_Declare(
+    nlohmann_json
+    URL https://github.com/nlohmann/json/archive/refs/tags/v3.11.2.tar.gz
+    URL_HASH MD5=e8d56bc54621037842ee9f0aeae27746
+  )
+  FetchContent_MakeAvailable(nlohmann_json)
+endif()
+set(CMAKE_CXX_STANDARD 17)
```
Presburger, line 94: no need set CMAKE_CXX_STANARD again.

CMakeLists.txt
```diff
@@ -75,10 +75,24 @@ include_directories(thirdparty/bitset)
 include_directories(thirdparty)
 
 find_package(OpenMP REQUIRED)
+
 if(WITH_RAFT)
   include(cmake/libs/libraft.cmake)
 endif()
 
+find_package(nlohmann_json)
```
Presburger, line 83: split this into cmake/lib as a nlohmann_json.cmake maybe looks better.

PR #734, opened 2023-03-13
adapt for raft ivf

Presburger:
1. Implemented serialization and deserialization of raft ivf index.
2. The thread works for thread pool, I hope one thread id mapping one gpu stream.
3. use the default device resource map, which can avoid some cuda error.

when I use python3 client, try the raft ivf index, old version got this error.
CUDA Error detected. cudaErrorCudartUnloading driver shutting down
seem some static variable destructor call's order cause this eror.
issue: #742


Presburger:
@wphicks hello, can you help me review my code, which adapt for raft ivf.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -37,6 +37,20 @@
 
 namespace knowhere {
 
+namespace memory_pool {
+static rmm::mr::device_memory_resource*
+resources() {
+    auto device_id = rmm::detail::current_device();
```
cjnolet, line 43: I propose we not invoke anything in the `detail` namespace (for RMM or for RAFT). That namespace is subject to change without warning and could cause integration nightmares going forward.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -158,15 +174,15 @@ class RaftIvfIndexNode : public IndexNode {
...
                     metric.value() != raft::distance::DistanceType::InnerProduct) {
                     LOG_KNOWHERE_WARNING_ << "selected metric not supported in RAFT IVF indexes: "
                                           << ivf_raft_cfg.metric_type;
                     return Status::invalid_metric_type;
                 }
-
+                devs_.insert(devs_.begin(), ivf_raft_cfg.gpu_ids.begin(), ivf_raft_cfg.gpu_ids.end());
                 auto scoped_device = detail::device_setter{*ivf_raft_cfg.gpu_ids.begin()};
-                res_ = std::make_unique<raft::device_resources>();
+                auto res_ = std::make_unique<raft::device_resources>(rmm::cuda_stream_per_thread, nullptr,
```
wphicks, line 184: Any reason we recreate the `device_resources` with every call? I'm assuming this is to ensure that we are getting the per-thread default stream each time.

I think we might be better served with a stream pool, which can be added directly to the `device_resources`. If we need to ensure that `Train` and `Add` complete before we perform a search, we can synchronize on the whole pool after those calls. Then, we can just pull streams from that pool to perform overlapping searches.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -373,11 +536,7 @@ class RaftIvfIndexNode : public IndexNode {
 
     virtual int64_t
     Count() const override {
-        auto result = std::int64_t{};
-        if (gpu_index_) {
-            result = gpu_index_->size();
-        }
-        return result;
+        return counts_;
```
wphicks, line 539: Is `Count` not supposed to return the total number of rows in the index? It looks like `counts_` is getting overridden with each `Add` call.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -344,12 +373,150 @@ class RaftIvfIndexNode : public IndexNode {
...
     virtual Status
     Serialize(BinarySet& binset) const override {
-        return Status::not_implemented;
+        if (!gpu_index_.has_value())
+            return Status::empty_index;
+        std::stringbuf buf;
+
+        std::ostream os(&buf);
+
+        os.write((char*)(&this->dim_), sizeof(this->dim_));
```
wphicks, line 382: This whole serialization setup seems fine, but I wonder if we wouldn't make our lives easier by upgrading to RAFT 23.04 and just invoking the [serialization API](https://github.com/rapidsai/raft/pull/1315) directly from that. It would likely dramatically simplify this logic. That can also happen as a follow up, if this is working for now.

wphicks (commented):
Thanks for tackling all this! I think we might want to tweak a few things. In particular, the way that RMM is being used now is a little unusual and may lead to some issues for us. I also think that we would probably be better served by just generating a stream pool on the `device_resources` handle rather than relying on the current workarounds to use default stream per thread.

src/index/ivf_raft/ivf_raft.cuh
cjnolet: Just to add to @wphicks, `device_resources` manages all of the resources that you need to interact w/ the GPU (and with other GPU-accelerated libraries like cublas) and they can be very expensive to create. As an example, reusing the same `device_resources` instead of creating a new one, specially during search, led to well over a 6x speedup in our recent 4-bit RAFT IVFPQ vs SCANN benchmarks. This matters a lot.

src/index/ivf_raft/ivf_raft.cuh
cjnolet: It would also centralize this so that, for example, you could load an index into Milvus that's been previously trained w/ RAFT.

src/index/ivf_raft/ivf_raft.cuh
Presburger: hello,  when we plug knowhere into milvus, milvus will check the count of the original dataset, I knowhere raft ivf will do some data padding, so, when we get the size from the raft index, milvus will crash down.

wphicks: Could you give us a little more detail on this? `index.size()` should return the actual size of the index without padding. Are you seeing a specific mismatch between the reported size and the number of added vectors? If you have reproducer code, that would be even better.

wphicks: My apologies! I just realized that what I said is true in 23.04 but false in 23.02 (which we are currently using. I'd recommend we upgrade for this and the other 23.04 features we've touched on elsewhere. I'm trying to get this PR to build right now, but once I've done so, I'll try to contribute code suggestions for how best to make use of 23.04.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -37,6 +37,58 @@
 
 namespace knowhere {
 
+namespace raft_res_pool {
+
+struct context {
+    static constexpr std::size_t default_size{16};
```
cjnolet, line 43: Is this intending to create 16 `rat::device_resources` instances up front in the same thread? All the `device_resources` instances are going to be using the same `cudaStream_t` underneath so it's not going to be asynchronous or otherwise parallelizable.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -158,15 +212,14 @@ class RaftIvfIndexNode : public IndexNode {
...
                     metric.value() != raft::distance::DistanceType::InnerProduct) {
                     LOG_KNOWHERE_WARNING_ << "selected metric not supported in RAFT IVF indexes: "
                                           << ivf_raft_cfg.metric_type;
                     return Status::invalid_metric_type;
                 }
-
+                devs_.insert(devs_.begin(), ivf_raft_cfg.gpu_ids.begin(), ivf_raft_cfg.gpu_ids.end());
                 auto scoped_device = detail::device_setter{*ivf_raft_cfg.gpu_ids.begin()};
-                res_ = std::make_unique<raft::device_resources>();
+                auto res_ = raft_res_pool::resource::instance().get_raft_res(devs_[0]);
```
cjnolet, line 222: I notice a stream-pool being created above. Is the goal to parallelize multiple queries and saturate the GPU by using different cuda streams or different threads? What guarantee does `Knowhere` make wrt the number of threads/gpus which might be invoking `search()` concurrently to compute the QPS?

If Knowhere is using different threads (and GPUs) to perform queries then we absolutely should be able to submit the queries on different streams to saturate each GPU. Though- I think we will want to be careful to make sure each GPU has its own copy of the IVF indexes in that case. 

We are going to need to set the stream from the stream pool explicitly on a copy of the `device_resources` instance for each stream in the stream pool, though. As it works right now, the `device_resources` can contain its own stream pool, which allows algorithms to use it internally when they support it. It's not automatic, though. If you are using `default_stream_per_thread` then I would create a new `device_resources` instance for each thread (maybe change `context()` to `thread_local` in that case?

wphicks:
Is it possible for us to populate the `device_resources` only on `Build` and `Load` calls and then assume that we have that available for each device at search time (rather than acquiring a lock and checking for existence on every call)? Given that we're benchmarking single-query calls, we should also assess the performance impact of using `std::map` to store the `device_resources` objects. If it is a significant part of the query time for small batches, shifting to a sorted vector would likely help us quite a bit.

Presburger:
> Is it possible for us to populate the `device_resources` only on `Build` and `Load` calls and then assume that we have that available for each device at search time (rather than acquiring a lock and checking for existence on every call)? Given that we're benchmarking single-query calls, we should also assess the performance impact of using `std::map` to store the `device_resources` objects. If it is a significant part of the query time for small batches, shifting to a sorted vector would likely help us quite a bit.

 This is a good proposal, we can init resource when build and load, and assume resource is available when we do search.
thanks

src/index/ivf_raft/ivf_raft.cuh
Presburger: no, we create a group of stream, std::vector<rmm::cuda_stream> stream_;

src/index/ivf_raft/ivf_raft.cuh
Presburger: knowhere have a thread pool, now search concurrent size is the cpu core number.

cjnolet: @Presburger okay so it sounds like multiple threads might be using the `raft::device_resources` resources but are all of the 
 the instances being created by the same thread?

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -37,6 +37,70 @@
...
+    context&
+    operator=(context&&) = delete;
+    context&
+    operator=(context const&) = delete;
+    raft::device_resources resources_;
+};
+
+context&
+get_context() {
+    thread_local context ctx;
```
cjnolet, line 64: This looks much better!

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -37,6 +37,70 @@
 
 namespace knowhere {
 
+namespace raft_res_pool {
+
+struct context {
+    context()
+        : resources_(
+              []() {
+                  return new rmm::cuda_stream();  // Avoid program exit datart
```
cjnolet, line 46: None of these arguments are actually needed. By default, the `device_resouces` instance will use the `per_thread_default_cuda_stream` and the `workspace_resources` (the last argument to the constructor here) will automatically use `rmm::mr::get_current_device_resource()`

wphicks (commented):
This looks great! In terms of the current error on exit, , it is happening because a static variable holds access to GPU memory, and the cuda context is destroyed before the static variable calls its destructor.

In the existing code, [these lines](https://github.com/milvus-io/knowhere/pull/734/files#diff-39396c81c69a6cd68c45dd364a208197954ee66f7086a2bb7853c1e441955a1eL129-L131) took care of that. In the new version, we just need to make sure that we call .release() on our pool allocators before we exit.

PR #740, opened 2023-03-15
Update QPS benchmark

cydrain:


benchmark/hdf5/benchmark_knowhere_float_qps.cpp
```diff
@@ -28,30 +28,30 @@ class Benchmark_knowhere_float_qps : public Benchmark_knowhere, public ::testing
...
                 if (recall < expected_recall) {
-                    left = nprobe;
+                    left = nprobe + 1;
                 } else {
-                    right = nprobe;
+                    right = nprobe - 1;
                 }
             }
-            return right;
+            return left;
```
PwzXxm, line 54: Should we return error here?

PR #751, opened 2023-03-17
Add Sanity Check after HNSW Build

PwzXxm:
This PR adds the functionality of performing a dummy search after the build process in HNSW.

This functionality should be used in the cloud environment for debugging purposes ONLY.

Catching Segmentation Faults and continuing to run the program can be tricky as signal and exceptions are two different systems in C++. We want to capture both in this case.

Be aware that this PR introduces compilation flags including `-fexceptions` and `-fnon-call-exceptions` to `knowhere` target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,6 +246,8 @@ if ( BUILD_COVERAGE STREQUAL "ON" )
                            )
 endif ()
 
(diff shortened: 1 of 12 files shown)


liliu-z (approved):
/lgtm

PR #752, opened 2023-03-18
Support GPU init in python test

cydrain:


include/knowhere/gpu/gpu_res_mgr.h
```diff
@@ -117,25 +121,22 @@ class GPUResMgr {
...
     GetRes() {
-        if (init_) {
-            auto res = res_bq_.Take();
-            return res;
-        } else {
-            KNOWHERE_THROW_MSG("GPUResMgr not initialized");
-        }
+        // Generally Init() should be called separately,
+        // here is for supporting python test
+        Init();
```
Presburger, line 126: check every time? init func have locks.

cydrain: after initialized, the lock will not be used any more

PR #753, opened 2023-03-18
Optimze conan better

Presburger:
issue: #779

### DIFF
--- a/benchmark/hdf5/benchmark_knowhere_float_qps.cpp
+++ b/benchmark/hdf5/benchmark_knowhere_float_qps.cpp
@@ -11,6 +11,7 @@
 
 #include <gtest/gtest.h>
 
+#include <thread>
 #include <vector>
 
 #include "benchmark_knowhere.h"
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,5 @@ python/knowhere/swigknowhere.py
 
 
 *.bin
+
+CMakeUserPresets.json
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,6 @@ include(cmake/libs/libhnsw.cmake)
 
 include_directories(thirdparty/faiss)
 include_directories(thirdparty/bitset)
-include_directories(thirdparty)
 
 find_package(OpenMP REQUIRED)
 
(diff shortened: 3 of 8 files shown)


benchmark/hdf5/benchmark_knowhere_float_qps.cpp
```diff
@@ -194,7 +194,12 @@ class Benchmark_knowhere_float_qps : public Benchmark_knowhere, public ::testing
 
 TEST_F(Benchmark_knowhere_float_qps, TEST_IVF_FLAT_NM) {
 #ifdef USE_CUDA
+#ifdef KNOWHERE_WITH_RAFT
+    index_type_ = knowhere::IndexEnum::INDEX_RAFT_IVFFLAT;
+#else
+
     index_type_ = knowhere::IndexEnum::INDEX_FAISS_GPU_IVFFLAT;
```
matrixji, line 201: Do we possibly do all benchmarks together, enable RFAT does not mean FAISS_GPU is not possible, and also enabling GPU not means CPU is not possible.
Maybe another PR could handle it.

Presburger: @matrixji in the future work, we will Abandon support for GPU in Faiss. raft looks better.

Presburger:
@matrixji  hello, I read you blog about the benchmark of faiss ivf & raft ivf, have you try the latest version of master? in my option, raft maybe much better than faiss when use multi thread query.

matrixji:
> @matrixji hello, I read you blog about the benchmark of faiss ivf & raft ivf, have you try the latest version of master? in my option, raft maybe much better than faiss when use multi thread query.

Not finished some tests with massive threads (which planned later). Currently, by reading some slides from GTC, I think RAFT is more suitable for online service, and the original faiss seems more suitable for batch work. And RAFT also has advantages in building an index, especially for a large index.

matrixji:
:-) Glad to see that hdf5 is resolved in cmake right now.

Presburger:
> > @matrixji hello, I read you blog about the benchmark of faiss ivf & raft ivf, have you try the latest version of master? in my option, raft maybe much better than faiss when use multi thread query.
> 
> Not finished some tests with massive threads (which planned later). Currently, by reading some slides from GTC, I think RAFT is more suitable for online service, and the original faiss seems more suitable for batch work. And RAFT also has advantages in building an index, especially for a large index.

yes, RAFT is better for milvus.

PR #757, opened 2023-03-21
Move gtest CMakeLists.txt to cmake/libs

cydrain:


Presburger:
gtest only used in benchmark, no need to move to cmake/libs

cydrain:
we are using conan already, need not this PR.

PR #759, opened 2023-03-21
remove omp process in diskann index

cqy123456:


thirdparty/DiskANN/src/partition_and_pq.cpp
```diff
@@ -493,7 +504,6 @@ int generate_pq_data_from_pivots(const std::string data_file,
     LOG_KNOWHERE_DEBUG_ << "Processing points  [" << start_id << ", " << end_id
                         << ")..";
 
-#pragma omp parallel for firstprivate(block_data_tmp) schedule(static, 8192)
```
liliu-z, line 496: Is it OK here to not replace omp with any pool?

liliu-z (commented):
1. Let's don't put this many codes with different purposes in a single PR.
2. Do we cleaned all OMP in DiskANN? I guess we still have some in Load path right?
3. Any performance change for this PR?

PR #766, opened 2023-03-22
Replace Easylogging++ by Glog

jiaoew1991:
issue: https://github.com/milvus-io/milvus/issues/22789
/kind improvement


liliu-z:
Ahaha Loves GLOG. Any chance to have some data about the performance comparison between them?

jiaoew1991:
> 

The API of `spdlog` is quite different from what milvus and knowhere are currently using, so many changes need to be made 😅

liliu-z:
> I'see raft depends spdlog(should be depends from rmm), so spdlog may also be a candidate. In my experience, spdlog has a similar interface to glog, and have better performance in multi-threads.

Good catch, and thanks for this suggestion.
For 3rdparty dependency, we will first keep what they are unless it affect the main project. From this point we don't have to align them.
For `spdlog`, thanks again for this proposal, and we can try to do some benchmark if time allows. If the result gives a strong support, we can definitely change all Milvus/Knowhere log system. All depends on cost&gain.

Always welcome suggestions

PR #788, opened 2023-03-29
Use folly threadpool

hhy3:
issue: #789 
Use folly threadpool to replace ctpl threadpool. This will improve performance a lot when qps is high

output


src/index/ivf_raft/ivf_raft.cu
```diff
@@ -14,36 +14,30 @@
...
 namespace knowhere {
 
-static std::shared_ptr<ThreadPool>
-GlobalThreadPoolRaft() {
-    static std::shared_ptr<ThreadPool> pool = std::make_shared<ThreadPool>(cuda_concurrent_size);
-    return pool;
-}
 KNOWHERE_REGISTER_GLOBAL(GPU_RAFT_IVF_FLAT, [](const Object& object) {
     return Index<IndexNodeThreadPoolWrapper>::Create(
-        std::make_unique<RaftIvfIndexNode<detail::raft_ivf_flat_index>>(object), GlobalThreadPoolRaft());
```
liliu-z, line 32: Why we removed all thread pools for RAFT?

hhy3: Because of the conflict between raft and folly, they can't be on the same compilation unit. So thread pool support for raft is moved to index_node_thread_pool_wrapper.cc

src/common/index_node_thread_pool_wrapper.cc
```diff
@@ -0,0 +1,50 @@
...
+#include "knowhere/index_node.h"
+
+namespace knowhere {
+
+namespace {
+
+constexpr uint32_t cuda_concurrent_size = 16;
+
+std::shared_ptr<ThreadPool>
+GlobalThreadPoolRaft() {
```
liliu-z, line 24: Super weird to have RAFT keyword here

liliu-z: Why you don't just create a constructor with num_thread(int) as input, and put this raft specific number into raft's code.

hhy3: Ok. Sounds reasonable

PR #791, opened 2023-03-31
Support GetVectorByIds for ANNOY, HNSW and all faiss index types

cydrain:


src/index/annoy/annoy.cc
```diff
@@ -139,23 +139,21 @@ class AnnoyIndexNode : public IndexNode {
 
         auto rows = dataset.GetRows();
         auto dim = dataset.GetDim();
-        auto p_ids = dataset.GetIds();
+        auto ids = dataset.GetIds();
 
-        float* p_x = nullptr;
+        float* data = new (std::nothrow) float[dim * rows];
```
PwzXxm, line 144: Why put `new` outside try block? If you use `std::nothrow`, it might be better to test for `nullptr` if it fails to allocate. Although it might throw an exception when accessing it, it is unclear where the exception comes from.

src/index/ivf/ivf.cc
```diff
@@ -469,25 +469,26 @@ IvfIndexNode<T>::GetVectorByIds(const DataSet& dataset, const Config& cfg) const
     }
     auto rows = dataset.GetRows();
     auto dim = dataset.GetDim();
-    float* p_x(new (std::nothrow) float[dim * rows]);
-    index_->make_direct_map(true);
-    auto p_ids = dataset.GetIds();
+    auto ids = dataset.GetIds();
+    float* data = nullptr;
     try {
+        data = new (std::nothrow) float[dim * rows];
```
PwzXxm, line 475: Might be better to be consistent with others

tests/ut/utils.h
```diff
@@ -42,6 +43,32 @@ GenDataSet(int rows, int dim, int seed = 42) {
     return ds;
 }
 
+inline std::unique_ptr<knowhere::DataSet>
+GenBinDataSet(int rows, int dim, int seed = 42) {
+    std::mt19937 rng(seed);
+    std::uniform_int_distribution<> distrib(0.0, 100.0);
```
PwzXxm, line 49: maybe better to use `std::numeric_limits<uint8_t>::min()` and `max()`?

src/index/annoy/annoy.cc
cydrain: updated

PR #796, opened 2023-04-03
Remove ANNOY index type

cydrain:


PwzXxm (changes requested):
Please remove annoy in `expected.h` and under `tests/python`

Issue #797, opened 2023-04-03
[Bug] DiskANN Failed to Query when Bitset is Set


PwzXxm:
Accessing invalid address when bitset is set.

PR #794, opened 2023-03-31
Use Brute Force when Bitset Filter Rate is High

PwzXxm:
Issue: #792 

This PR contains:
1. Use brute force strategy when bitset filter rate is high on `HNSW` and `DiskANN`
2. In UT, set `nq` from `1000` to `100` for CPU indices.
3. In UT, test for recall instead of searching for base vectors themselves.

PR #798, opened 2023-04-03
Fix DiskANN Access Invalid Address when Query with Bitset

PwzXxm:
issue: #797 

When bitset is set, when `cur_list_size` is `1` and the bitset of the only item in the `retset` is set, it got `memmove`d and `cur_list_size` set to `0`. Then when try to insert to the `retset` later, accessing `right = cur_list_size - 1` will be invalid.

Related test cases will be introduced in the PR #794


hhy3 (approved):
/lgtm

PR #807, opened 2023-04-10
Rename Timer to time_recorder

cydrain:


PwzXxm:
Checked with @cydrain that timer messages are not parsed by tests so far.

Issue #808, opened 2023-04-11
Knowhere didn't maintain the Topk limitation same for every index.


xiaofan-luan:
If user tune the topk limit on milvus, the behaviour of current knowhere 1.x is weird:
1. for faiss, it looks ok and will just accept the top K
2. for HNSW.  ef needs to be larger than topK limit and smaller than HNSW_MAX_EF. since HNSW_MAX_EF is a fixed number, topk is actually fixed to be less than HNSW_MAX_EF but we did't have ensure it.
3. for diskANN, I didn't find anywhere limit topk, is there any where?

xiaofan-luan:
what happens in knowhere 2.0 if topk is larger than ef? 
I don't find any limitation on it

cydrain:
Hi @cqy123456 , can you help to check the topk limitation in Diskann

PR #809, opened 2023-04-11
Change the Ef limit for HNSW when topk is large

xiaofan-luan:
see also #808


liliu-z:
/lgtm
/approve

Quick question, do we need to modify this in other indexes?

PR #810, opened 2023-04-12
Add mdspan API to raft IVF functions

lowener:
This PR changes the raft ivf function calls from raw pointer to mdspan. These functions are pre-compiled in libraft so it will speed up the compile time.

To receive the latest libraft release, it is recommended to use conda package manager:
```
conda install -c rapidsai -c nvidia -c rapidsai-nightly libraft==23.04.00a
```
I'm open for suggestions on how to handle installing conda and linking to libraft.
issue: #857


Presburger:
> This PR changes the raft ivf function calls from raw pointer to mdspan. These functions are pre-compiled in libraft so it will speed up the compile time.
> 
> To receive the latest libraft release, it is recommended to use conda package manager:
> 
> ```
> conda install -c rapidsai -c nvidia -c rapidsai-nightly libraft==23.04.00a
> ```
> 
> I'm open for suggestions on how to handle installing conda and linking to libraft.

thanks for contribute the code,
is it heavily dependent on to the conda?

lowener:
Conda is the preferred way to get libraft usually. Since conda is not being used right now I found an alternative with pip through the pylibraft package.

Presburger:
@lowener Essentially, we are using knowhere more as a c++ library, and if he needs to rely on the python environment, it will be troublesome for us in the vast majority of cases.

Presburger:
@lowener This pr is excellent and I would accept it if you remove the HINT_LIBRAFT related code. thank you.

Presburger (approved):

PR #812, opened 2023-04-12
Add metric type COSINE

cydrain:


src/common/utils.cc
```diff
@@ -0,0 +1,61 @@
...
+
+namespace knowhere {
+
+const float floatDiff = 0.00001;
+
+void
+normalize(float* data, int32_t dim) {
+    double vecLen = 0.0;
+    for (int32_t j = 0; j < dim; j++) {
+        vecLen += data[j] * data[j];
```
liliu-z, line 28: float overflow risky, maybe we can do `a += (double)x * (double)c`

src/common/utils.cc
```diff
@@ -0,0 +1,61 @@
...
+
+#include "knowhere/comp/time_recorder.h"
+#include "knowhere/log.h"
+
+namespace knowhere {
+
+const float floatDiff = 0.00001;
+
+void
+normalize(float* data, int32_t dim) {
```
liliu-z, line 25: Pretty confusing with this signature. Since it claims to normalize vectors (because we have dim) without rows.
Suggestions:
1. Change the dim to num.
2. Change the normalize to NormalizeVec (or some other names), and data to vector.
3. Add a new parameter called row and default it by 1. For loop normalize `data` as a list of vectors.
Let's pick one of the ways to fix it.

include/knowhere/comp/brute_force.h
```diff
@@ -20,14 +20,15 @@ namespace knowhere {
 class BruteForce {
  public:
     static expected<DataSetPtr, Status>
-    Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset, const Json& config, const BitsetView& bitset);
+    Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset, const Json& json, const BitsetView& bitset);
```
liliu-z, line 23: Let's keep it config. Its type is Json but role here is config, so its variable name should be config.

include/knowhere/index.h
```diff
@@ -116,6 +117,7 @@ class Index {
     Status
     Build(const DataSet& dataset, const Json& json) {
         Json json_(json);
+        normalize(dataset, json_);
```
liliu-z, line 120: Let's put the COSINE check here in index.h, since if not COSINE we shouldn't even call a `normalize` function. Also, if the purpose is just to centralize the code, we can have a private inline function for it here.

liliu-z: Also, use `IsMetricType` here, that is a good practice.

liliu-z (commented):
Comment on the hacky in-place modification plz.

src/common/utils.cc
cydrain: updated

include/knowhere/index.h
cydrain: updated

src/common/utils.cc
cydrain: updated

xiaofan-luan:
would this be backport to Milvus 2.2.X?

liliu-z:
> would this be backport to Milvus 2.2.X?

In current plan, we will just support it after Milvus 2.3. Plz let us know if it is necessary to have it in Milvus 2.2.x.

include/knowhere/utils.h
```diff
@@ -0,0 +1,30 @@
...
+
+#pragma once
+
+#include "knowhere/config.h"
+#include "knowhere/dataset.h"
+
+namespace knowhere {
+
+inline bool
+IsMetricType(const std::string& str, const knowhere::MetricType& metric_type) {
```
liliu-z, line 20: To my understanding, only `brute_force` need string comparison for metrics and all others can directly use enum matching right?

Can we have a new API to compare enum class instead of turning them to string.

cydrain: only after metric string converted to faiss::metric, we can use enum matching,
for HNSW and DISKANN, we have to use string compare.

PR #813, opened 2023-04-13
Remove HNSW param ef invalid check from ut

cydrain:


liliu-z:
/kind improvement
/approve

Do we have these kinds of ut for HNSW's k, and DiskANN's k/search list size?

Issue #818, opened 2023-04-14
Build Knowhere Python Wheels on Multiple Platforms and Python Versions


PwzXxm:
Platforms to support:
* Linux x86_64
* Linux aarch64
* MacOS

Python Versions to support:
- 3.8
- 3.9
- 3.10
- 3.11

PwzXxm:
Support Linux x86_64 with the Python version above for now.

Building Linux aarch64 on GitHub Action Runners via QEMU emulators may take several hours.

PR #819, opened 2023-04-14
Build Python Wheels on manylinux x86 and Upload to PyPI

PwzXxm:
issue: #818 

Use `cibuildwheels` to build wheels on different platforms for the Knowhere Python package. The `auditwheel` tool in the `cibuildwheels` will copy libraries in the shared library to reach maximum portability possible.


PwzXxm:
For python wheels release action, check https://github.com/PwzXxm/knowhere/actions/runs/4696182428

scripts/python_deps.sh
```diff
@@ -0,0 +1,62 @@
...
+    fi
+fi
+
+if [[ "${MACHINE}" == "Mac"  ]]; then
+    brew install libomp openblas
+    pip3 install conan==1.58.0
+
+    echo "Setting CC and CXX to use clang"
+    export CC=/usr/local/opt/llvm/bin/clang
+    export CXX=/usr/local/opt/llvm/bin/clang++
```
Presburger, line 61: why need use clang?

PwzXxm: use `llvm-clang` rather than the default Xcode clang. Xcode clang toolchain is broken with `libomp`

PR #822, opened 2023-04-14
raft index, fix param m and add env for setting gpu mem pool size

Presburger:
issue: #826

### DIFF
--- a/src/common/comp/knowhere_config.cc
+++ b/src/common/comp/knowhere_config.cc
@@ -21,7 +21,9 @@
 #include "index/gpu/gpu_res_mgr.h"
 #endif
 #include "simd/hook.h"
-
+#ifdef KNOWHERE_WITH_RAFT
+#include "index/ivf_raft/set_pool.h"
+#endif
 namespace knowhere {
 
 void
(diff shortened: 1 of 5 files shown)


src/common/comp/knowhere_config.cc
```diff
@@ -147,5 +149,11 @@ KnowhereConfig::FreeGPUResource() {
     knowhere::GPUResMgr::GetInstance().Free();
 #endif
 }
+void
+KnowhereConfig::SetRaftMemPool(size_t init_size, size_t max_size) {
+#ifdef KNOWHERE_WITH_RAFT
+    knowhere::SetRaftMemPool(init_size, max_size);
```
liliu-z, line 155: Can we not compile `KnowhereConfig::SetRaftMemPool` if RAFT is not enable instead of just skipping calling `knowhere::SetRaftMemPool`.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -540,5 +552,11 @@ class RaftIvfIndexNode : public IndexNode {
     std::optional<T> gpu_index_;
 };
 
+void
+SetRaftMemPool(size_t init_size, size_t max_size) {
```
liliu-z, line 556: Who's the caller?

Presburger: call from knowhere_config, milvus will call it.

Issue #792, opened 2023-03-31
Use Brute Force when Bitset Filter Rate is High


PwzXxm:

PwzXxm:
Waiting for experiment results to set threshold

PR #829, opened 2023-04-21
Update HNSW Brute Force Threshold While Searching with Bitset

PwzXxm:
issue: #792 

Based on the experiment results below, when the filter rate is higher than ~93%, brute force outperforms graph search at around ~1M vector scale from GIST and GLOVE datasets.

Thanks @cydrain for running the experiments.

image
image


liliu-z (approved):
/lgtm

PR #831, opened 2023-04-22
Use PQ and Refine for DiskANN when Bitset Filter Ratio is High

PwzXxm:
issue: #792 

Calculate PQ distances and refine, instead of brute force strategy when bitset filter ratio is high, to reduce the number of I/Os performed.

The threshold where switching the strategy can be affected by multiple factors, such as topk, dimension, distribution of the dataset and etc. We fit a cautious log function regarding topk and having a safety net setting at 0.5.

This boosts the performance by 80x in the best case.

gist_final


liliu-z (approved):
/lgtm

PR #836, opened 2023-04-24
Fix diskann COSINE support

cydrain:


cydrain:
> thanks cydrain! Is there a ganruatee that all of the index in milvus support cosine, including: Faiss, HNSW, DiskANN and AutoIndex

Yes, all float index types can support COSINE now

Issue #844, opened 2023-04-26
Show "DEBUG" in knowhere version when build debug image


cydrain:
Show "DEBUG" in knowhere version when build debug image

PR #860, opened 2023-05-06
Show DEBUG in knowhere version for debug build

cydrain:
Issue: #844

### DIFF
--- a/src/common/comp/knowhere_config.cc
+++ b/src/common/comp/knowhere_config.cc
@@ -24,21 +24,30 @@
 #ifdef KNOWHERE_WITH_RAFT
 #include "index/ivf_raft/set_pool.h"
 #endif
+
 namespace knowhere {
 
 void
 KnowhereConfig::ShowVersion() {
 #define XSTR(x) STR(x)
 #define STR(x) #x
+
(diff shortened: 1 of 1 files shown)


cydrain:
test with Milvus debug build, can see following message:
```
I20230506 11:35:00.246528 51480 knowhere_config.cc:47] [KNOWHERE][ShowVersion][milvus] Knowhere Version: 6448cf09 (DEBUG)
```

src/common/comp/knowhere_config.cc
```diff
@@ -24,21 +24,27 @@
...
-#else
-    LOG_KNOWHERE_INFO_ << "Knowhere Version: " << XSTR(KNOWHERE_VERSION);
+    msg = msg + "-gpu";
 #endif
 #else
-    LOG_KNOWHERE_INFO_ << "Knowhere Version: unknown";
+    msg = msg + " unknown";
+#endif
+
+#ifdef _DEBUG
```
PwzXxm, line 44: AFAIK, `_DEBUG` is not in the standard, but `NDEBUG` is. Can you double-check?

cydrain: ok, updated

PR #865, opened 2023-05-10
Update get vector ut

cydrain:


tests/ut/test_get_vector.cc
```diff
@@ -17,25 +17,81 @@
...
     int64_t nb = 10000;
     int64_t dim = 128;
     int64_t seed = 42;
 
-    auto base_gen = [&]() {
+    auto base_bin_gen = [&]() {
         knowhere::Json json;
         json[knowhere::meta::DIM] = dim;
-        json[knowhere::meta::METRIC_TYPE] = knowhere::metric::L2;
+        json[knowhere::meta::METRIC_TYPE] = knowhere::metric::HAMMING;
```
PwzXxm, line 30: how about other binary metrics?

tests/ut/test_get_vector.cc
```diff
@@ -17,25 +17,81 @@
...
+}
+
+TEST_CASE("Test Float Get Vector By Ids", "[Float GetVectorByIds]") {
+    using Catch::Approx;
+
+    int64_t nb = 10000;
+    int64_t dim = 128;
+    int64_t seed = 42;
+
+    auto metric = GENERATE(as<std::string>{}, knowhere::metric::L2, knowhere::metric::COSINE);
```
PwzXxm, line 89: how about IP?

tests/ut/test_get_vector.cc
cydrain: for binary index types, metric type does not matter

tests/ut/test_get_vector.cc
cydrain: for float index, when COSINE metric type, get vector may return not-original vector.
so it's enough to only test L2 and COSINE

Issue #859, opened 2023-05-06
[bug]:[recall] IVFPQ index rangesearch precision reduction 


elstic:
index: IVFPQ
data_type: glove
case  :
`pytest test_ivf_pq.py::TestIvf::test_ivf_recall`

```
[2023-05-05 16:57:11 - INFO - root]: ********** IVFPQ recall:simd:avx2, data_type:glove, nlist:1024 ********** (test_ivf_pq.py:33) (140118922323776:MainThread)
[2023-05-05 16:57:11 - INFO - root]: hdf5 dataset: <HDF5 file "glove-200-angular.hdf5" (mode r)> (test_ivf_pq.py:37) (140118922323776:MainThread)
[2023-05-05 16:57:19 - INFO - root]: data entities: 1183514, metric type:IP, latency_errs:0.5 (test_ivf_pq.py:47) (140118922323776:MainThread)
[2023-05-05 16:58:05 - INFO - root]: IVFPQ build time: 45.6555s @-0.6455 (test_ivf_pq.py:75) (140118922323776:MainThread)
[2023-05-05 16:58:07 - INFO - root]: nprobe=1, nq=10000, topk=100, search in 0.142s @-0.3545, recall=0.1353 (test_ivf_pq.py:110) (140118922323776:MainThread)
[2023-05-05 16:58:08 - INFO - root]: nprobe=2, nq=10000, topk=100, search in 0.1304s @-0.0383, recall=0.1549 (test_ivf_pq.py:110) (140118922323776:MainThread)
[2023-05-05 16:58:10 - INFO - root]: nprobe=4, nq=10000, topk=100, search in 0.1861s @-0.3516, recall=0.1656 (test_ivf_pq.py:110) (140118922323776:MainThread)
[2023-05-05 16:58:12 - INFO - root]: nprobe=8, nq=10000, topk=100, search in 0.2572s @-0.2556, recall=0.1705 (test_ivf_pq.py:110) (140118922323776:MainThread)
[2023-05-05 16:58:14 - INFO - root]: nprobe=16, nq=10000, topk=100, search in 0.4211s @-0.1628, recall=0.1733 (test_ivf_pq.py:110) (140118922323776:MainThread)
[2023-05-05 16:58:16 - INFO - root]: nprobe=32, nq=10000, topk=100, search in 0.6892s @-0.2411, recall=0.175 (test_ivf_pq.py:110) (140118922323776:MainThread)
[...]

Issue #858, opened 2023-05-06
[bug]:[recall] BIN_IVF_FLAT index recall lowering


elstic:
index:  BIN_IVF_FLAT
dataset: jaccard

```
********** BIN_IVF_FLAT recall:simd:avx2, data_type:sift-4096-jaccard, nlist:1024 ********** (test_bin_ivf.py:31) (140118922323776:MainThread)
[2023-05-05 15:24:11 - INFO - root]: hdf5 dataset: <HDF5 file "sift-4096-jaccard.hdf5" (mode r)> (test_bin_ivf.py:35) (140118922323776:MainThread)
[2023-05-05 15:24:15 - INFO - root]: data entities: 1000000, metric type:JACCARD, latency_errs:0.5 (test_bin_ivf.py:45) (140118922323776:MainThread)
[2023-05-05 15:25:55 - INFO - root]: BIN_IVF_FLAT build time: 99.4018s @-0.7132 (test_bin_ivf.py:70) (140118922323776:MainThread)
[2023-05-05 15:25:57 - INFO - root]: nprobe=1, nq=10000, topk=100, search in 1.113s @-0.5147, recall=0.0869 (test_bin_ivf.py:107) (140118922323776:MainThread)
[2023-05-05 15:26:00 - ERROR - root]: nprobe2 recall degraded: expected=0.1469, actual=0.1455 (test_bin_ivf.py:97) (140118922323776:MainThread)
[2023-05-05 15:26:01 - INFO - root]: nprobe=2, nq=10000, topk=100, search in 1.9632s @-0.5342, recall=0.1455 (test_bin_ivf.py:107) (140118922323776:MainThread)
[2023-05-05 15:26:05 - ERROR - root]: nprobe4 recall degraded: expected=0.2326, actual=0.2308 (test_bin_ivf.py:97) (140118922323776:MainThread)
[2023-05-05 15:26:06 - INFO - root]: nprobe=4, nq=10000, topk=100, search in 3.5794s @-0.5443, recall=0.2308 (test_bin_ivf.py:107) (140118922323776:MainThread)
[2023-05-05 15:26:13 - ERROR - root]: nprobe8 recall degraded: expected=0.3478, actual=0.3447 (test_bin_ivf.py:97) (140118922323776:MainThread)
[...]

PR #867, opened 2023-05-10
Roll back #842 to fix IVF_PQ and BIN_IVF_FLAT recall degrade

cydrain:
Issue: #858 #859


PwzXxm:
Rolled back AVX for normalization, using SSE where possible

PR #869, opened 2023-05-10
Add cosine for diskann

hhy3:
There is no significant difference in performance.
fuck

### DIFF
--- a/thirdparty/DiskANN/include/diskann/utils.h
+++ b/thirdparty/DiskANN/include/diskann/utils.h
@@ -768,6 +768,78 @@ namespace diskann {
     return max_norm;
   }
 
+  template<typename T>
+  std::vector<float> prepare_base_for_cosine(const std::string in_file,
+                                        const std::string out_file) {
+    LOG_KNOWHERE_DEBUG_
+        << "Pre-processing base file by normalizing";
+    std::ifstream in_reader(in_file.c_str(), std::ios::binary);
+    std::ofstream out_writer(out_file.c_str(), std::ios::binary);
+    _u64          npts, in_dims, out_dims;
+
+    _u32 npts32, dims32;
+    in_reader.read((char*) &npts32, sizeof(uint32_t));
+    in_reader.read((char*) &dims32, sizeof(uint32_t));
+
+    npts = npts32;
+    in_dims = dims32;
+    out_dims = in_dims;
+    _u32 outdims32 = (_u32) out_dims;
+
+    out_writer.write((char*) &npts32, sizeof(uint32_t));
+    out_writer.write((char*) &outdims32, sizeof(uint32_t));
+
+    size_t               BLOCK_SIZE = 100000;
+    size_t               block_size = npts <= BLOCK_SIZE ? npts : BLOCK_SIZE;
+    std::unique_ptr<T[]> in_block_data =
+        std::make_unique<T[]>(block_size * in_dims);
+    std::unique_ptr<float[]> out_block_data =
+        std::make_unique<float[]>(block_size * out_dims);
+
+    std::memset(out_block_data.get(), 0, sizeof(float) * block_size * out_dims);
+    _u64 num_blocks = DIV_ROUND_UP(npts, block_size);
+
+    std::vector<float> norms(npts, 0);
+
+    for (_u64 b = 0; b < num_blocks; b++) {
+      _u64 start_id = b * block_size;
+      _u64 end_id = (b + 1) * block_size < npts ? (b + 1) * block_size : npts;
+      _u64 block_pts = end_id - start_id;
+      in_reader.read((char*) in_block_data.get(),
+                     block_pts * in_dims * sizeof(T));
+      for (_u64 p = 0; p < block_pts; p++) {
+        for (_u64 j = 0; j < in_dims; j++) {
+          norms[start_id + p] +=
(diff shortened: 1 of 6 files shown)


thirdparty/DiskANN/include/diskann/utils.h
```diff
@@ -768,6 +768,76 @@ namespace diskann {
...
+    for (_u64 b = 0; b < num_blocks; b++) {
+      _u64 start_id = b * block_size;
+      _u64 end_id = (b + 1) * block_size < npts ? (b + 1) * block_size : npts;
+      _u64 block_pts = end_id - start_id;
+      in_reader.read((char*) in_block_data.get(),
+                     block_pts * in_dims * sizeof(T));
+      for (_u64 p = 0; p < block_pts; p++) {
+        for (_u64 j = 0; j < in_dims; j++) {
+          norms[start_id + p] +=
+              in_block_data[p * in_dims + j] * in_block_data[p * in_dims + j];
```
liliu-z, line 813: Should check the float overflow here?

thirdparty/DiskANN/include/diskann/utils.h
```diff
@@ -768,6 +768,76 @@ namespace diskann {
...
+
+    npts = npts32;
+    in_dims = dims32;
+    out_dims = in_dims;
+    _u32 outdims32 = (_u32) out_dims;
+
+    out_writer.write((char*) &npts32, sizeof(uint32_t));
+    out_writer.write((char*) &outdims32, sizeof(uint32_t));
+
+    size_t               BLOCK_SIZE = 100000;
```
liliu-z, line 792: Can we align this number with buffer we use in other places during the building process to keep the memory usage steady

thirdparty/DiskANN/include/diskann/utils.h
```diff
@@ -768,6 +768,76 @@ namespace diskann {
...
+        std::make_unique<T[]>(block_size * in_dims);
+    std::unique_ptr<float[]> out_block_data =
+        std::make_unique<float[]>(block_size * out_dims);
+
+    std::memset(out_block_data.get(), 0, sizeof(float) * block_size * out_dims);
+    _u64 num_blocks = DIV_ROUND_UP(npts, block_size);
+
+    std::vector<float> norms(npts, 0);
+
+    for (_u64 b = 0; b < num_blocks; b++) {
```
liliu-z, line 804: Should we do parallel here?

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -1297,10 +1309,13 @@ namespace diskann {
         if (bitset_view.empty() || !bitset_view.test(frontier_nhood.first)) {
           float cur_expanded_dist;
           if (!use_disk_index_pq) {
+            if (metric == Metric::COSINE) {
+              normalize(node_fp_coords_copy, aligned_dim);
+            }
```
liliu-z, line 1314: Looks like our logic is doing normalization for the base vector for every single search. Will this be time consuming? As discussed, we planned to store the norm of all base vectors during the building time to skip this part right?

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -264,6 +264,9 @@ namespace diskann {
         T    *node_coords = OFFSET_TO_NODE_COORDS(node_buf);
         T    *cached_coords = coord_cache_buf + node_idx * aligned_dim;
         memcpy(cached_coords, node_coords, disk_bytes_per_point);
+        if (metric == diskann::Metric::COSINE) {
+          normalize(cached_coords, aligned_dim);
+        }
```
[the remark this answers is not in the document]
hhy3: Ok

thirdparty/DiskANN/include/diskann/utils.h
hhy3: Ok

thirdparty/DiskANN/include/diskann/utils.h
hhy3: It's aligned with line 718 above.

thirdparty/DiskANN/src/pq_flash_index.cpp
hhy3: By benchmarking this will not cause performance problem

thirdparty/DiskANN/include/diskann/utils.h
hhy3: IMO there's no need to parallel here cause it's not cpu bound

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -801,6 +806,13 @@ namespace diskann {
                           << this->max_base_norm;
       delete[] norm_val;
```
liliu-z, line 807: need return after this

thirdparty/DiskANN/include/diskann/utils.h
hhy3: Which number are you referring to

thirdparty/DiskANN/src/pq_flash_index.cpp
hhy3: why return is needed after this

liliu-z (approved):
/lgtm
/approve

PR #873, opened 2023-05-12
Replace LoadConfig with Config

yah01:
/kind improvement

### DIFF
--- a/include/knowhere/config.h
+++ b/include/knowhere/config.h
@@ -56,11 +56,12 @@ template <typename T>
 struct Entry {};
 
 enum PARAM_TYPE {
-    TRAIN = 0x1,
-    SEARCH = 0x2,
-    RANGE_SEARCH = 0x4,
-    FEDER = 0x8,
-    DESERIALIZE = 0x10,
+    TRAIN = 1 << 0,
+    SEARCH = 1 << 1,
+    RANGE_SEARCH = 1 << 2,
+    FEDER = 1 << 3,
+    DESERIALIZE = 1 << 4,
+    DESERIALIZE_FROM_FILE = 1 << 5,
 };
 
 template <>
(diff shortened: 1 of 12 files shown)


include/knowhere/config.h
```diff
@@ -222,11 +224,15 @@ class EntryAccess {
...
+        entry->type |= PARAM_TYPE::LOAD;
+        return *this;
+    }
+
     EntryAccess&
     for_all() {
-        entry->type |= PARAM_TYPE::TRAIN;
-        entry->type |= PARAM_TYPE::SEARCH;
-        entry->type |= PARAM_TYPE::RANGE_SEARCH;
+        entry->type |= PARAM_TYPE::ALL;
```
liliu-z, line 235: 1. I know the naming is confusing, but looks like this `for_all` is only for `train`, `search` and `range search`, not for `feder`. So this change will mess it up.
2. Considering the forward compatibility, let's don't add a enum `ALL` here, or it is easy to trap the following modifier. For instance, I am about to adding one enum for deserialization, and without noticing there is an `ALL`, it is easy to make a mistake.

include/knowhere/config.h
```diff
@@ -56,10 +56,12 @@ template <typename T>
...
 enum PARAM_TYPE {
-    TRAIN = 0x1,
-    SEARCH = 0x2,
-    RANGE_SEARCH = 0x4,
-    FEDER = 0x8,
+    TRAIN = 1 << 0,
+    SEARCH = 1 << 1,
+    RANGE_SEARCH = 1 << 2,
+    FEDER = 1 << 3,
+    LOAD = 1 << 4,
```
liliu-z, line 63: Let's follow the API name, do DESERIALIZE_FROM_FILE

liliu-z (changes requested):

PR #874, opened 2023-05-12
Update index BinarySet key name to be compatible with knowhere-1.x

cydrain:


src/index/flat/flat.cc
```diff
@@ -265,10 +265,12 @@ class FlatIndexNode : public IndexNode {
             }
             std::shared_ptr<uint8_t[]> data(writer.data_);
             if constexpr (std::is_same<T, faiss::IndexFlat>::value) {
-                binset.Append("FLAT", data, writer.rp);
+                // use "IVF" to be compatible with knowhere-1.x
+                binset.Append("IVF", data, writer.rp);
```
liliu-z, line 269: For backward compatibility we can support both `FLAT` and `IVF` as Flat Index for `Deserialization` and only support `FLAT` for Serialization

liliu-z (changes requested):

src/index/flat/flat.cc
cydrain: updated

PR #875, opened 2023-05-12
optimize gpu bitset filter

Presburger:
issue: #852

### DIFF
--- a/src/index/ivf_raft/ivf_raft.cuh
+++ b/src/index/ivf_raft/ivf_raft.cuh
@@ -45,42 +45,86 @@
 namespace knowhere {
 
 __global__ void
-filter(const int k1, const int k2, const int nq, const uint8_t* bs, int64_t* ids_before, float* dis_before,
-       int64_t* ids, float* dis) {
+filter(const uint8_t* bs, const int64_t* ids_before, int32_t* ids_block, int len) {
+    int tx = blockIdx.x * blockDim.x + threadIdx.x;
+    if (tx >= len)
+        return;
+    __shared__ int64_t ids_before_s[1024];
+    __shared__ int32_t ids_block_s[1024];
+
+    ids_before_s[threadIdx.x] = ids_before[tx];
+    int64_t i = ids_before_s[threadIdx.x];
+    ids_block_s[threadIdx.x] = 1 - (bs[i >> 3] && (0x1 << (i & 0x7)));
+    ids_block[tx] = ids_block_s[threadIdx.x];
+}
+
+__global__ void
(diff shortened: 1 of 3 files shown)


src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -517,9 +570,17 @@ class RaftIvfIndexNode : public IndexNode {
...
+                        auto ids_block = raft::make_device_vector<int32_t, std::int64_t>(*res_, rows * k2);
+                        filter<<<(rows * k2 + 1023) / 1024, 1024, 0, stream.value()>>>(
+                            bs_gpu.data_handle(), ids_gpu_before.data_handle(), ids_block.data_handle(), rows * k2);
+                        auto pre_ids_block = raft::make_device_vector<int32_t, std::int64_t>(*res_, rows * k2);
+                        RAFT_CUDA_TRY(cudaMemcpyAsync(pre_ids_block.data_handle(), ids_block.data_handle(),
+                                                      rows * k2 * sizeof(int32_t), cudaMemcpyDefault, stream.value()));
+                        prescan<<<dim3(1, rows), k2 / 2, sizeof(int32_t) * k2, stream.value()>>>(
+                            pre_ids_block.data_handle(), ids_block.data_handle(), k2);
+                        write_back<<<dim3(1, rows), k1, sizeof(int32_t) * k2, stream.value()>>>(
+                            pre_ids_block.data_handle(), ids_gpu_before.data_handle(), dis_gpu_before.data_handle(),
```
foxspy, line 582: k1 should be limited because threads number could not exceed 1024 for Cuda. Explicitly check this param in config if necessary.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -45,42 +45,86 @@
...
+            left = mid + 1;
+        }
+    }
+    if (left < k2) {
+        ids_after[ty * k1 + tx] = ids_before[ty * k2 + left];
+        dis_after[ty * k1 + tx] = dis_before[ty * k2 + left];
+    } else {
+        ids_after[ty * k1 + tx] = -1;
+        dis_after[ty * k1 + tx] = 1.0f / 0.0f;
+        ids_before[0] = -1;  // destory old ans;
```
foxspy, line 126: In my opinion, a better way is just scan k2 data and write back into k1 result. This way use k1 result to find data from k2 data, which may cause bad performance when bitset is dense.

Presburger: Good Advise.

foxspy (approved):
/lgtm

PR #879, opened 2023-05-15
Add c api for knowhere, therefor other language like Python&Rust can use it easily

jiaoew1991:


liliu-z:
Opened a new branch call `with_c_api`. Feel free to check in that one.

PR #882, opened 2023-05-16
Update knowhere_config.cc

SleepyBear96:
enum SimdType {
        AUTO = 0,  // enable all and depend on the system
        AVX512,    // only enable AVX512
        AVX2,      // only enable AVX2
        SSE4_2,    // only enable SSE4_2
        GENERIC,   // use arithmetic instead of SIMD
    };

### DIFF
--- a/src/common/comp/knowhere_config.cc
+++ b/src/common/comp/knowhere_config.cc
@@ -60,8 +60,8 @@ KnowhereConfig::SetSimdType(const SimdType simd_type) {
         LOG_KNOWHERE_INFO_ << "FAISS expect simdType::AUTO";
     } else if (simd_type == SimdType::AVX512) {
         faiss::use_avx512 = true;
-        faiss::use_avx2 = true;
-        faiss::use_sse4_2 = true;
+        faiss::use_avx2 = false;
+        faiss::use_sse4_2 = false;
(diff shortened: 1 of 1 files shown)


cydrain:
Hi @SleepyBear96 ,
Why need this change ?

SleepyBear96:
> Hi @SleepyBear96 , Why need this change ?

The definition of the SimdType says, AVX512 only enable AVX512

cydrain:
> > Hi @SleepyBear96 , Why need this change ?
> 
> The definition of the SimdType says, AVX512 only enable AVX512



> > Hi @SleepyBear96 , Why need this change ?
> 
> The definition of the SimdType says, AVX512 only enable AVX512



> > Hi @SleepyBear96 , Why need this change ?
> 
> The definition of the SimdType says, AVX512 only enable AVX512

Hi @SleepyBear96,

You can see the code in src/simd/hook.cc  Line 75
avx512 APIs are hooked with 2 conditions:
1. user required: it's controlled by variable "faiss::use_avx512"
2. hardware supported: we use API cpu_support_avx512() to decide whether avx512 can be supported

Here we set all of "use_avx512", "use_avx2" and "use_sse4_2" to true, is because if cpu_support_avx512() return false,
we will use avx2 instead.

PR #886, opened 2023-05-17
fix raft compile error in milvus and optize bitset filter

Presburger:
issue: #852

### DIFF
--- a/src/index/ivf_raft/ivf_raft.cuh
+++ b/src/index/ivf_raft/ivf_raft.cuh
@@ -54,7 +54,11 @@ filter(const uint8_t* bs, const int64_t* ids_before, int32_t* ids_block, int len
 
     ids_before_s[threadIdx.x] = ids_before[tx];
     int64_t i = ids_before_s[threadIdx.x];
-    ids_block_s[threadIdx.x] = 1 - (bs[i >> 3] && (0x1 << (i & 0x7)));
+    if (i != -1)
+        ids_block_s[threadIdx.x] = 1 - (bs[i >> 3] && (0x1 << (i & 0x7)));
+    else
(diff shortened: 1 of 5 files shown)


src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -123,10 +127,21 @@ write_back(const int32_t* pre_ids_block, int64_t* ids_before, float* dis_before,
     } else {
         ids_after[ty * k1 + tx] = -1;
         dis_after[ty * k1 + tx] = 1.0f / 0.0f;
-        ids_before[0] = -1;  // destory old ans;
+        ids_before[0] = -2;  // destory old ans;
```
foxspy, line 130: Use a single variable could be better.

src/index/ivf_raft/ivf_raft_config.h
```diff
@@ -64,6 +70,12 @@ class RaftIvfPqConfig : public IvfPqConfig {
     CFG_STRING internal_distance_dtype;
     CFG_FLOAT preferred_shmem_carveout;
     KNOHWERE_DECLARE_CONFIG(RaftIvfPqConfig) {
+        KNOWHERE_CONFIG_DECLARE_FIELD(k)
+            .set_default(10)
+            .description("search for top k similar vector.")
+            .set_range(1, 1024)
+            .for_search();
+
```
foxspy, line 78: The upper bound value 1024 is limited by hardware. However, this limitation can be easily broken by optimizing software, which can reduce the explanation cost.

Presburger: Good Idead.

foxspy (approved):
/lgtm

PR #887, opened 2023-05-17
Let Deserialize() return empty_binary_set when key invalid

cydrain:


src/index/flat/flat.cc
```diff
@@ -278,6 +278,9 @@ class FlatIndexNode : public IndexNode {
                                           "BinaryIVF",  // compatible with knowhere-1.x
                                           Type()};
         auto binary = binset.GetByNames(names);
+        if (binary == nullptr) {
+            return Status::invalid_binary_set;
+        }
```
liliu-z, line 283: When will this happen?

PR #892, opened 2023-05-19
Fix diskann ip low recall for unnormalized data

foxspy:
image
Fixed
image
image
Fixed by adding the 2 times of the product of the added dim.


thirdparty/DiskANN/src/index.cpp
```diff
@@ -226,7 +226,7 @@ namespace diskann {
   }
 
   template<typename T, typename TagT>
-  Index<T, TagT>::Index(Metric m, const size_t dim, const size_t max_points,
+  Index<T, TagT>::Index(Metric m, bool ip_prepared, const size_t dim, const size_t max_points,
```
liliu-z, line 229: Can I ask why we don'y simply use `Metrics m` as the indicator?

foxspy: image
metric is L2 all the time even for IP.

PR #902, opened 2023-05-24
Add some logs

hhy3:


include/knowhere/config.h
```diff
@@ -341,13 +358,15 @@ class Config {
...
                 if (json.find(it.first) == json.end() && !ptr->default_val.has_value()) {
+                    LOG_KNOWHERE_ERROR_ << "invalid param [" << it.first << "] in json.";
                     return Status::invalid_param_in_json;
                 }
                 if (json.find(it.first) == json.end()) {
                     *ptr->val = ptr->default_val.value();
                     continue;
                 }
                 if (!json[it.first].is_array()) {
+                    LOG_KNOWHERE_ERROR_ << "Type conflict in json: param [" << it.first << "] should be a string.";
```
cydrain, line 369: should be an array

src/common/comp/brute_force.cc
```diff
@@ -117,8 +118,10 @@ BruteForce::Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset
                                            cur_distances, cur_labels, bitset);
                     break;
                 }
-                default:
+                default: {
+                    LOG_KNOWHERE_ERROR_ << "Invalid metric type [" << cfg.metric_type << "].";
```
cydrain, line 122: suggest to change like this:
LOG_KNOWHERE_ERROR_ << "Invalid metric type: " << cfg.metric_type;

PR #913, opened 2023-05-29
Add some prometheus metrics

cydrain:


include/knowhere/expected.h
```diff
@@ -112,4 +112,20 @@ class expected {
 
 }  // namespace knowhere
 
+#define CHECK_STATUS(X)                         \
```
PwzXxm, line 115: Maybe change these to `KNOWHERE_CHECK_STATUS` to avoid definition pollution.

include/knowhere/index.h
```diff
@@ -16,6 +16,11 @@
 #include "knowhere/index_node.h"
 #include "knowhere/log.h"
 
+#ifdef NOT_COMPILE_FOR_SWIG
```
PwzXxm, line 19: Didn't see it defined/used/documented anywhere in this PR.

include/knowhere/index.h
```diff
@@ -117,90 +122,84 @@ class Index {
...
-        }
-        return this->node->Search(dataset, *cfg, bitset);
+        CHECK_EXPECTED(Config::FormatAndCheck(*cfg, json_));
+        LOG_KNOWHERE_DEBUG_ << "Search config dump: " << json_.dump();
+        CHECK_EXPECTED(Config::Load(*cfg, json_, knowhere::SEARCH));
+
+#ifdef NOT_COMPILE_FOR_SWIG
+        TimeRecorder rc("Search");
+        auto res = this->node->Search(dataset, *cfg, bitset);
+        auto span = rc.ElapseFromBegin("done");
```
PwzXxm, line 176: Does this function print any message? Excessive messages may slow down search process

include/knowhere/index.h
cydrain: This MACRO "NOT_COMPILE_FOR_SWIG" is defined in knowhere/CMakeLists.txt Line 62
The code embraced by this MACRO will not be compiled by python wheel.
Using this MACRO here is because python wheel need not compile prometheus metrics.

include/knowhere/expected.h
cydrain: ok

include/knowhere/index.h
cydrain: TimeRecorder is generated with "log_level_ = 0", it will print nothing

PwzXxm:
/lgtm
Performance tested with subtle differences by @cydrain

include/knowhere/index.h
```diff
@@ -117,90 +122,84 @@ class Index {
...
-        auto res = Config::Load(*cfg, json_, knowhere::RANGE_SEARCH);
-        if (res != Status::success) {
-            return unexpected(res);
-        }
-        return this->node->RangeSearch(dataset, *cfg, bitset);
+        KNOWHERE_CHECK_EXPECTED(Config::FormatAndCheck(*cfg, json_));
+        LOG_KNOWHERE_DEBUG_ << "RangeSearch config dump: " << json_.dump();
+        KNOWHERE_CHECK_EXPECTED(Config::Load(*cfg, json_, knowhere::RANGE_SEARCH));
+
+#ifdef NOT_COMPILE_FOR_SWIG
```
liliuleo, line 193: Why need to have this

zhengbuqian: To avoid compiling prometheus for python interface, which is now problematic as of I know

PR #915, opened 2023-06-05
Refactored expected class to remove unexpected and simplify usage

zhengbuqian:
1. Make the template parameter of `expected` to be `<T>` instead of `<T, E>`: `expected` is designed to contain either a success value or a fail status, thus no need to make `Status` a template parameter. I don't see `E` to be other classes in current code and I don't aforesee any future other `E` candidates.
2. Remove `unexpected` class and allow `expected` to be constructed directly from `Status`. So we can do just `return Status::some_error:` instead of `return unexpected(Status:some_error);`;
3. Added 2 macros for easier error status handling(avoids tons of boilerplate code):

```cpp
// Functions that return `expected<T>`
// without this pr:
auto res = CallSomeFunc();
if (!res.has_value()) {
  LOG(...);
  return res.error();
}
auto res_value = res.value();

// with this pr:
ASSIGN_OR_RETURN(T, res_value, LOG(...), CallSomeFunc());

// Functions that return Status
// without this pr:
Status res = CallSomeFunc();
if (res != Status::success) {
  return res;
}

// with this pr:
RETURN_IF_ERROR(CallSomeFunc());
```

4. slightly refactored `index.h` to extract some duplicate codes into a method
5. in `diskann.cc` make `TryDiskANNCall` non-template and make its argument function have no return value, which is clearer.

This PR contains no feature/performance change.

### DIFF
--- a/include/knowhere/expected.h
+++ b/include/knowhere/expected.h
@@ -41,91 +41,114 @@ enum class Status {
     invalid_binary_set = 19,
 };
 
-template <typename E>
-class unexpected {
- public:
-    constexpr explicit unexpected(const E& err) : err(err) {
-        static_assert(std::is_same<E, Status>::value);
-    }
-
-    constexpr explicit unexpected(E&& err) : err(err) {
-        static_assert(std::is_same<E, Status>::value);
-    }
-
-    ~unexpected() = default;
-
-    E err;
-};
-
-template <typename T, typename E>
+template <typename T>
 class expected {
  public:
     template <typename... Args>
-    expected(Args&&... args) : val(std::forward<Args...>(args...)) {
+    expected(Args&&... args) : val(std::make_optional<T>(std::forward<Args>(args)...)) {
     }
 
-    expected(const unexpected<E>& unexp) {
-        err = unexp.err;
+    expected(const Status& err) : err(err) {
+        assert(err != Status::success);
     }
 
-    expected(unexpected<E>&& unexp) {
-        err = std::move(unexp.err);
+    expected(Status&& err) : err(err) {
+        assert(err != Status::success);
     }
 
-    expected(const expected<T, E>&) = default;
+    expected(const expected<T>&) = default;
 
-    expected(expected<T, E>&&) noexcept = default;
+    expected(expected<T>&&) noexcept = default;
 
     expected&
-    operator=(const expected<T, E>&) = default;
+    operator=(const expected<T>&) = default;
 
     expected&
-    operator=(expected<T, E>&&) noexcept = default;
+    operator=(expected<T>&&) noexcept = default;
 
     bool
-    has_value() {
+    has_value() const {
         return val.has_value();
     }
 
-    E
+    Status
     error() const {
         assert(val.has_value() == false);
         return err.value();
     }
 
-    T&
-    value() {
+    const T&
+    value() const {
         assert(val.has_value() == true);
         return val.value();
     }
 
-    expected<T, E>&
-    operator=(const unexpected<E>& unexp) {
-        err = unexp.err;
+    expected<T>&
+    operator=(const Status& err) {
+        assert(err != Status::success);
+        this->err = err;
         return *this;
     }
 
  private:
     std::optional<T> val = std::nullopt;
-    std::optional<E> err = std::nullopt;
+    std::optional<Status> err = std::nullopt;
 };
 
+// Evaluates expr that returns a Status. Does nothing if the returned Status is
+// a Status::success, otherwise returns the Status from the current function.
+#define RETURN_IF_ERROR(expr)            \
+    do {                                 \
+        auto status = (expr);            \
+        if (status != Status::success) { \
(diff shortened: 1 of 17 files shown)


include/knowhere/index.h
```diff
@@ -284,6 +261,15 @@ class Index {
         static_assert(std::is_base_of<IndexNode, T1>::value);
     }
 
+    Status
+    LoadConfig(BaseConfig* cfg, const Json& json, knowhere::PARAM_TYPE param_type, std::string method) const {
```
cydrain, line 265: use std::string&

src/index/ivf/ivf.cc
```diff
@@ -197,12 +197,8 @@ IvfIndexNode<T>::Build(const DataSet& dataset, const Config& cfg) {
     const BaseConfig& base_cfg = static_cast<const IvfConfig&>(cfg);
     auto build_thread_num = base_cfg.get_build_thread_num();
     ThreadPool::ScopedOmpSetter setter(build_thread_num);
-    auto err = Train(dataset, cfg);
-    if (err != Status::success) {
-        return err;
-    }
-    err = Add(dataset, cfg);
-    return err;
+    RETURN_IF_ERROR(Train(dataset, cfg));
```
cydrain, line 200: suggest to change like this:
    RETURN_IF_ERROR(Train(dataset, cfg));
    RETURN_IF_ERROR(Add(dataset, cfg));
    return Status::success;

src/index/hnsw/hnsw.cc
```diff
@@ -38,12 +38,8 @@ class HnswIndexNode : public IndexNode {
 
     Status
     Build(const DataSet& dataset, const Config& cfg) override {
-        auto res = this->Train(dataset, cfg);
-        if (res != Status::success) {
-            return res;
-        }
-        res = Add(dataset, cfg);
-        return res;
+        RETURN_IF_ERROR(Train(dataset, cfg));
```
cydrain, line 41: suggest to change like this:
        RETURN_IF_ERROR(Train(dataset, cfg));
        RETURN_IF_ERROR(Add(dataset, cfg));
        return Status::success;

src/index/diskann/diskann.cc
```diff
@@ -380,10 +376,12 @@ DiskANNIndexNode<T>::Deserialize(const BinarySet& binset, const Config& cfg) {
 
     pq_flash_index_ = std::make_unique<diskann::PQFlashIndex<T>>(reader, diskann_metric);
 
-    auto load_expect =
-        TryDiskANNCall<int>([&]() -> int { return pq_flash_index_->load(pool_->size(), index_prefix_.c_str()); });
-
-    if (!load_expect.has_value() || load_expect.value() != 0) {
+    if (TryDiskANNCall([&]() {
```
cydrain, line 379: suggest not change this code

src/common/comp/brute_force.cc
```diff
@@ -152,22 +141,16 @@ BruteForce::SearchWithBuf(const DataSetPtr base_dataset, const DataSetPtr query_
...
-        return Status::invalid_metric_type;
-    }
+    RETURN_IF_ERROR(Config::Load(cfg, config, knowhere::SEARCH));
 
     int topk = cfg.k;
     auto labels = ids;
     auto distances = dis;
 
-    auto faiss_metric_type = metric_type.value();
+    ASSIGN_OR_RETURN(faiss::MetricType, faiss_metric_type,
```
cydrain, line 150: if there is LOG, suggest keep the code as it be

include/knowhere/expected.h
```diff
@@ -99,33 +83,72 @@ class expected {
...
+    do {                                 \
+        auto status = (expr);            \
+        if (status != Status::success) { \
+            return status;               \
+        }                                \
+    } while (0)
+
+template <typename T>
+Status
+DoAssignOrReturn(T& lhs, expected<T> exp) {
```
cydrain, line 110: use "DoAssignOrReturn(T& lhs, const expected<T>& exp) "

zhengbuqian: Done

src/common/comp/brute_force.cc
zhengbuqian: Done, removed logging from ASSIGN_OR_RETURN

src/index/diskann/diskann.cc
zhengbuqian: Chatted offline, extracted the lambda into be a variable for better readability

src/index/hnsw/hnsw.cc
zhengbuqian: ditto

src/index/ivf/ivf.cc
zhengbuqian: Chatted offline, agreed to keep it as is

include/knowhere/index.h
zhengbuqian: Done

PR #922, opened 2023-06-08
Refactor GPU post-filtering logic

### LINKED ISSUE #928: Refactor custom GPU kernels for greater flexibility and robustness
As currently implemented, the GPU kernels used for postprocessing and filtering results from RAFT are launched with a data-dependent number of threads and blocks. For performance and robustness, it is generally advisable to use [grid-stride loops](https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/) and similar techniques such that algorithmic correctness is independent of those parameters. This ensures that total thread and block counts do not exceed hardware limits and also allows tuning of the block and thread count to achieve optimal performance.

While running benchmarks with the existing implementation, I also observed GPU memory access errors. Refactoring these kernels eliminated those memory access issues, though I did not determine the specific lines of the existing implementation that caused those problems.

See PR #922.

wphicks:
This PR refactors the GPU post-filtering logic to make it a little more modular and robust. Bitset related logic is now encapsulated in a bitset object that can be used on-device. Postprocessing kernels are also written such that correctness is independent of the kernel launch parameters.

Resolve #928


environment.yml
```diff
@@ -0,0 +1,18 @@
+---
```
wphicks, line 1: This file should be removed prior to merger. Added as a convenience for collaborating within our team on these changes.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -45,103 +49,111 @@
...
-            int32_t t = temp[ai];
-            temp[ai] = temp[bi];
-            temp[bi] += t;
+
+        // Now move all valid results to the front of the row
+        auto cur_valid_index = row_index * k;
+        for (auto col_index = 0; col_index < k; ++col_index) {
+            auto elem_index = row_index * k + col_index;
+            // Just do one row per block for now; can improve this later
+            if (ids[elem_index] != -1 && threadIdx.x == 0) {
```
cjnolet, line 123: With a batched block-level cumsum (treating invalid as 0's), you might be able to get each thread in the block independently writing to their assigned output locations.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -435,7 +447,7 @@ class RaftIvfIndexNode : public IndexNode {
                                               cudaMemcpyDefault, stream.value()));
 
                 auto indices = rmm::device_uvector<std::int64_t>(rows, stream);
-                thrust::sequence(thrust::device, indices.begin(), indices.end(), gpu_index_->size());
+                thrust::sequence(thrust::device.on(stream.value()), indices.begin(), indices.end(), gpu_index_->size());
```
cjnolet, line 450: Nitpit: Any reason we can't use `res.get_thrust_policy()` here?

src/index/ivf_raft/ivf_raft.cuh
cjnolet: I realized my prior comment wasn't explained very well. I just mean treating each thread as a "k" position where it stores 0 if invalid and 1 if valid and doing a cumsum of the resulting mask.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -45,103 +49,111 @@
...
+    rows() const {
+        return ids_.extent(0);
+    }
+    auto
+    k() const {
+        return ids_.extent(1);
+    }
+
+ private:
+    raft::device_matrix<std::int64_t, std::int64_t> ids_;
```
cjnolet, line 89: Nice to see these being used more!

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -568,63 +532,24 @@ class RaftIvfIndexNode : public IndexNode {
...
             }
-            RAFT_CUDA_TRY(cudaMemcpyAsync(ids.get(), ids_gpu.data_handle(), ids_gpu.size() * sizeof(std::int64_t),
-                                          cudaMemcpyDefault, stream.value()));
-            RAFT_CUDA_TRY(cudaMemcpyAsync(dis.get(), dis_gpu.data_handle(), dis_gpu.size() * sizeof(float),
-                                          cudaMemcpyDefault, stream.value()));
-            stream.synchronize();
+
+            if (gpu_results.k() != ivf_raft_cfg.k) {
+                auto new_gpu_results = raft_detail::raft_results{*res_, gpu_results.rows(), ivf_raft_cfg.k};
+                raft_detail::slice<<<1024, 256, 0, res_->get_stream().value()>>>(
```
cjnolet, line 543: Just curious- where did the choice of 1024 for the number of blocks come from? This could probably be chosen based on the number of SMs.

src/index/ivf_raft/ivf_raft.cuh
wphicks: No reason at all! Will update accordingly, thanks!

src/index/ivf_raft/ivf_raft.cuh
wphicks: Just a safe default for all architectures we're building for. Given the time constraints, I didn't want to do the necessary device querying to get the optimal configuration. I don't expect this to be a bottleneck anyway just yet, and there are a bunch of other performance tweaks that will probably be even higher impact. That being said, we should definitely improve on this in subsequent iterations.

src/index/ivf_raft/ivf_raft.cuh
wphicks: Oh that's a great thought! Do the block-level cumsum and then that gives us non-overlapping pairs of input/output locations that each thread can work on swapping. I'm going to try to get some perf numbers on this tomorrow, and if they're weak (and this is a hotspot), that will be one of the first things I try. Very nice idea!

cjnolet: Yeah I was about to say if the perf isn't bottlenecked here then there might not even be a point in doing this at all. The idea just popped up as I was looking over the loop in the kernel.

wphicks:
This still needs tests to cover the previously failing cases, but the core logic is ready for review. I will work on tests tomorrow. I will also try to get a Milvus build working with this so we can get perf numbers and see if we should tackle any low-hanging fruit for optimizations before release.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -45,103 +49,111 @@
...
-    while (left < right) {
-        int mid = (left + right) / 2;
-        if (temp[mid] == tx + 1) {
-            right = mid;
-        } else if (temp[mid] > tx + 1) {
-            right = mid;
-        } else if (temp[mid] < tx + 1) {
-            left = mid + 1;
+slice(T* out, T* in, IndexT out_rows, IndexT out_cols, IndexT in_rows, IndexT in_cols) {
+    for (auto i = blockIdx.x * blockDim.x + threadIdx.x; i < out_rows * out_cols; ++i) {
```
lowener, line 145: Is this for-loop condition set correctly? Thread #0 will do `out_rows * out_cols` iteration and Thread #1 will do `out_rows * out_cols - 1` iteration?

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -45,103 +49,111 @@
...
+        auto row_index = i / out_cols;
+        auto col_index = i % out_cols;
+        if (row_index < in_rows && col_index < in_cols) {
+            out[row_index * out_cols + col_index] = in[row_index * in_cols + col_index];
```
lowener, line 149:
```suggestion
            out[i] = in[row_index * in_cols + col_index];
```

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -45,103 +49,111 @@
...
+        } else {
+            out[row_index * out_cols + col_index] = -1;
```
lowener, line 151:
```suggestion
            out[i] = -1;
```

src/index/ivf_raft/ivf_raft.cuh
wphicks: Whoops! Can't believe I screwed up a grid-strided loop, sigh...

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -20,12 +20,15 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <iostream>
```
wphicks, line 23: Note to self: I think this can be removed.

src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -45,103 +49,111 @@
...
-    int right = k2;
-    while (left < right) {
-        int mid = (left + right) / 2;
-        if (temp[mid] == tx + 1) {
-            right = mid;
-        } else if (temp[mid] > tx + 1) {
-            right = mid;
-        } else if (temp[mid] < tx + 1) {
-            left = mid + 1;
+slice(T* out, T* in, IndexT out_rows, IndexT out_cols, IndexT in_rows, IndexT in_cols) {
```
wphicks, line 144: Note to self: This needs a comment explaining its use.

src/index/ivf_raft/ivf_raft.cuh
Presburger: > With a batched block-level cumsum (treating invalid as 0's), you might be able to get each thread in the block independently writing to their assigned output locations. Not super important for the first iteration here but just an idea for future optimizations.

It seems that our previous implementation was along these linesi

wphicks:
I should mention for the purposes of review that we can probably make these kernels a bit more performant in a later release, but they are fairly speedy as is and do not seem to be the bottleneck for search.

PR #923, opened 2023-06-09
Set search_list(ef) default size and throw exception if search list(ef) < topk

cqy123456:
Set search_list(ef) default size and throw exception if search list(ef) < topk
issue: #910

### DIFF
--- a/knowhere/index/vector_index/IndexHNSW.cpp
+++ b/knowhere/index/vector_index/IndexHNSW.cpp
@@ -29,6 +29,21 @@
 #include "index/vector_index/helpers/RangeUtil.h"
 
 namespace knowhere {
+namespace {
+    inline int64_t
+    CheckAndGetEfValue(const Config& config) {
+        auto topk_val = GetMetaTopk(config);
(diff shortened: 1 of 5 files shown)


knowhere/index/vector_index/IndexHNSW.cpp
```diff
@@ -29,6 +29,21 @@
 #include "index/vector_index/helpers/RangeUtil.h"
 
 namespace knowhere {
+namespace {
+    inline int64_t
+    CheckAndGetEfValue(const Config& config) {
+        auto topk_val = GetMetaTopk(config);
+        if (CheckKeyInConfig(config, indexparam::EF)) {
```
cydrain, line 36: suggest move these code to HNSWConfAdapter::CheckSearch
        if (CheckKeyInConfig(config, indexparam::EF)) {
            auto ef_val = GetIndexParamEf(config);
            if (ef_val < topk_val) {
                KNOWHERE_THROW_MSG("ef is smaller than topk in hnsw.");
            }
        }

PR #926, opened 2023-06-13
Use a empty json string as json default value in python api

cqy123456:
issue: #927


PwzXxm:
DiskANN deserialization uses a JSON config, and other index types need a default value.

PR #933, opened 2023-06-14
Refactor GPU filtering logic for greater robustness

### LINKED ISSUE #928: Refactor custom GPU kernels for greater flexibility and robustness

wphicks:
This PR refactors the GPU post-filtering logic to make it a little more modular and robust. Bitset related logic is now encapsulated in a bitset object that can be used on-device. Postprocessing kernels are also written such that correctness is independent of the kernel launch parameters.

Resolve #928

### DIFF
--- a/include/knowhere/device_bitset.h
+++ b/include/knowhere/device_bitset.h
@@ -1,0 +1,90 @@
+// Copyright (C) 2023 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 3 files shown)


wphicks:
This PR replaces #922, squashing all changes into a single signed-off commit.

Presburger (approved):

PR #934, opened 2023-06-15
Add benchmark_knowhere_float_range_bitset.cpp

cydrain:
HNSW range search with bitset
Screenshot from 2023-06-15 17-00-15

HNSW bruteforce range search with bitset
Screenshot from 2023-06-15 17-00-34

HNSW range search with bitset
Screenshot from 2023-06-15 19-46-35

HNSW bruteforce range search with bitset
Screenshot from 2023-06-15 19-46-55


cydrain:
> Should we introduce another threshold for range search separately?

ok, already use another threshold for range search

Issue #919, opened 2023-06-06
Diskann GetVectorByIds latency not stable


cydrain:
Diskann GetVectorByIds latency not stable, can reproduce with following patch

branch: main
commit: 2fdbe7f4
[diskann_getvector.diff.txt](https://github.com/milvus-io/knowhere/files/11685561/diskann_getvector.diff.txt)

build: make test-cpp
run: cmake_build/bin/all_tests --gtest_filter="*IndexGetVector*"

cydrain:
![Screenshot from 2023-06-08 16-11-48](https://github.com/milvus-io/knowhere/assets/3992404/5d14fc78-d925-4b27-b974-094db78229b3)

PwzXxm:
Decreased around 23% of total time and 50% when concurrency was high. It should now perform much better when concurrency is high or the number of requested IDs is large.

The fluctuations of I/O latency are considered normal to NVMe SSD. See PR #935

PR #935, opened 2023-06-15
Improve DiskANN GetVectorByIds Performance

PwzXxm:
issue: #919 

Before the change, `GetVectorByIds` in DiskANN deduplicate ids according to its sector offsets and fire a single Knowhere thread per SSD sector read per libaio request.

After the change, it divides ids into batches and fires threads on each batch, these batches then being deduplicated and fire a single libaio request per batch.

It also takes care about vectors with a large dimension.

## E2E Milvus Results
Get 50 ids, measure the time of 200 iterations, DiskANN cache off
Initial runs are ignored

  | Before (s) | After (s)
-- | -- | --
Concurrency 1 | 0.151025 | 0.1153
Concurrency 4 | 0.431405 | 0.292123
Concurrency 20 | 1.86118 | 0.962817

Decreased around 23% of total time and 50% when concurrency was high. It should now perform much better when concurrency is high or the number of requested IDs is large.

### DIFF
--- a/src/index/diskann/diskann.cc
+++ b/src/index/diskann/diskann.cc
@@ -169,6 +169,7 @@ namespace knowhere {
 namespace {
 static constexpr float kCacheExpansionRate = 1.2;
 static constexpr int kSearchListSizeMaxValue = 200;
+static constexpr int64_t kGetVectorBatchSize = 32;
 template <typename T>
 expected<T, Status>
 TryDiskANNCall(std::function<T()>&& diskann_call) {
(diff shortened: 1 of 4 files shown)


PwzXxm:
Vec same as before, test script without sleep(300ms)
Get 50 ids, measure the time of 200 iterations, DiskANN cache off
Initial runs are ignored

  | Before (s) | After (s)
-- | -- | --
Concurrency 1 | 0.151025 | 0.1153
Concurrency 4 | 0.431405 | 0.292123
Concurrency 20 | 1.86118 | 0.962817

Decreased around 23% of total time and 50% when concurrency was high. It should now perform much better when concurrency is high or the number of requested IDs is large. 

The fluctuations of I/O latency are considered normal to NVMe SSD.

thirdparty/DiskANN/src/pq_flash_index.cpp
```diff
@@ -1527,31 +1528,57 @@ namespace diskann {
...
   template<typename T>
-  void PQFlashIndex<T>::get_vector_by_sector(const _u64 sector_offset,
-                                             const std::vector<_u64> &ids_idx,
-                                             const int64_t           *ids,
-                                             T *output_data) {
+  void PQFlashIndex<T>::get_vector_by_ids(const int64_t *ids, const int64_t n, T *output_data) {
+    size_t batch_size = kReadBatchSize;
+    if (long_node) {
+      auto min_size = kReadBatchSize / nsectors_per_node;
+      batch_size = (min_size == 0) ? (MAX_N_SECTOR_READS * read_len_for_node / nsectors_per_node) : min_size;
```
cqy123456, line 1535: sizeof(sector_scratch ) = MAX_N_SECTOR_READS * read_len_for_node 
                                        = MAX_N_SECTOR_READS * (SECTOR_LEN * nsectors_per_node) 
sizeof(sector_scratch )  > = read_len_for_node * batch_size  
means :
MAX_N_SECTOR_READS > batch_size ?

PwzXxm: Indeed

src/index/diskann/diskann.cc
```diff
@@ -169,6 +169,7 @@ namespace knowhere {
 namespace {
 static constexpr float kCacheExpansionRate = 1.2;
 static constexpr int kSearchListSizeMaxValue = 200;
+static constexpr int64_t kGetVectorBatchSize = 32;
```
xiaofan-luan, line 172: Why do we believe 32 is a good

PwzXxm: Updated in PR #962

PR #936, opened 2023-06-15
Sync HNSW ef behavior with knowhere-1.x

cydrain:
ef is set:
  ef >= topk ==> pass
  ef < topk   ==> err out

ef not set:
  set ef to topk

### DIFF
--- a/include/knowhere/config.h
+++ b/include/knowhere/config.h
@@ -475,6 +475,11 @@ class BaseConfig : public Config {
         }
         return omp_get_max_threads();
     }
+
+    virtual Status
+    CheckAndAdjustConfig() {
+        return Status::success;
+    }
 };
 }  // namespace knowhere
 
--- a/include/knowhere/index.h
+++ b/include/knowhere/index.h
@@ -169,6 +169,7 @@ class Index {
         KNOWHERE_CHECK_EXPECTED(Config::FormatAndCheck(*cfg, json_));
         LOG_KNOWHERE_DEBUG_ << "Search config dump: " << json_.dump();
         KNOWHERE_CHECK_EXPECTED(Config::Load(*cfg, json_, knowhere::SEARCH));
+        KNOWHERE_CHECK_EXPECTED(cfg->CheckAndAdjustConfig());
 
 #ifdef NOT_COMPILE_FOR_SWIG
         TimeRecorder rc("Search");
--- a/src/index/hnsw/hnsw_config.h
+++ b/src/index/hnsw/hnsw_config.h
@@ -12,7 +12,12 @@
 #ifndef HNSW_CONFIG_H
 #define HNSW_CONFIG_H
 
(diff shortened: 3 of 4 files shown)


include/knowhere/index.h
```diff
@@ -169,6 +169,7 @@ class Index {
         KNOWHERE_CHECK_EXPECTED(Config::FormatAndCheck(*cfg, json_));
         LOG_KNOWHERE_DEBUG_ << "Search config dump: " << json_.dump();
         KNOWHERE_CHECK_EXPECTED(Config::Load(*cfg, json_, knowhere::SEARCH));
+        KNOWHERE_CHECK_EXPECTED(cfg->Customize(json));
```
liliu-z, line 172: To my understanding. After loading, the config object is loaded, can we use it directly instead the json?

include/knowhere/config.h
```diff
@@ -473,6 +473,11 @@ class BaseConfig : public Config {
         }
         return omp_get_max_threads();
     }
+
+    virtual Status
+    Customize(const Json& json) {
```
liliu-z, line 478: This function name express no useful info about what it does. Let's rename it to `IsValid` or `CheckValidation` something like this

liliu-z (changes requested):

include/knowhere/index.h
cydrain: I need use the input json to tell whether ef is set by user or by default value

include/knowhere/config.h
cydrain: this API is not a readonly API, it may modify the value in the Config.

include/knowhere/index.h
liliu-z: Then this will too confusing since we should not put it after loading.

As mentioned, we don't need to set default value for `ef`, then we can check the config here.

include/knowhere/config.h
liliu-z: `CheckAndAdjustConfig` sounds better to me then, since we need to let people know what is this API doing.

cydrain: rename API to `CheckAndAdjustConfig`

src/index/hnsw/hnsw_config.h
```diff
@@ -12,8 +12,11 @@
 #ifndef HNSW_CONFIG_H
 #define HNSW_CONFIG_H
 
+#include "knowhere/comp/index_param.h"
 #include "knowhere/config.h"
 
+const int64_t defaultHnswEf = -1;
```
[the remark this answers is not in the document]
liliu-z: And also comments is necessary for this strategy. Thanks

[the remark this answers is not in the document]
cydrain: updated

liliu-z (approved):
/lgtm

PR #951, opened 2023-06-23
Merge Train/Add to Build

liliu-z:
Knowhere has `Build`, `Train` and `Add` APIs, their semantic should be:
`Build`: build a index in a batch.
`Train`: train the index w/o adding data
`Add`: add data to indexes.

For DiskANN, we should not support `Train` & `Add` but only `Build`

### DIFF
--- a/src/index/diskann/diskann.cc
+++ b/src/index/diskann/diskann.cc
@@ -43,17 +43,17 @@ class DiskANNIndexNode : public IndexNode {
     }
 
     Status
-    Build(const DataSet& dataset, const Config& cfg) override {
-        return Add(dataset, cfg);
-    }
+    Build(const DataSet& dataset, const Config& cfg) override;
 
     Status
     Train(const DataSet& dataset, const Config& cfg) override {
-        return Status::success;
+        return Status::not_implemented;
     }
 
     Status
-    Add(const DataSet& dataset, const Config& cfg) override;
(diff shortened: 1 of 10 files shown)


foxspy (approved):
/lgtm

src/index/diskann/diskann.cc
```diff
@@ -43,17 +43,17 @@ class DiskANNIndexNode : public IndexNode {
...
     Status
-    Build(const DataSet& dataset, const Config& cfg) override {
-        return Add(dataset, cfg);
-    }
+    Build(const DataSet& dataset, const Config& cfg) override;
 
     Status
     Train(const DataSet& dataset, const Config& cfg) override {
-        return Status::success;
+        return Status::not_implemented;
```
zhengbuqian, line 50: not implemented error should be used when either the code is being developed and it not implemented yet; or the base class forces its subclasses to implement a specific method by providing a default impl to return not_implemented error. `not_implemented` methods should eventually be implemented either after development or by subclasses.

IIUC we never want the Train/Add methods of DiskANN to be invoked, not_implemented doesn't seem to be the correct error code. How about invalid_args or a new code Status::failed_precondition/not_supported?

liliu-z: Good catch, free feel to make a PR adding that

PR #953, opened 2023-06-24
Refactor device resource management to control memory usage

### LINKED ISSUE #952: Stream-per-thread leads to excessive device memory usage
Currently, in an environment (such as a Milvus server) where knowhere GPU searches are launched from many host threads, we observe excessive device memory usage. This is because each host thread can independently launch its own searches, resulting in separate simultaneous GPU allocations for each thread.

To ensure that GPUs do not run out of memory for large indexes, host threads should share a limited number of streams per device.

wphicks:
This PR provides more streamlined handling of device resource management that also allows configuration of the number of CUDA streams per device. This ensures that GPU indexes will not exhaust device memory on deployments with many host threads.

This PR also lays the groundwork for multi-device usage via replication of the index across devices, but it does not actually add that capability yet.

Resolve #952


src/index/ivf_raft/ivf_raft.cuh
```diff
@@ -651,8 +627,15 @@ class RaftIvfIndexNode : public IndexNode {
 
 void
 SetRaftMemPool(size_t init_size, size_t max_size) {
-    LOG_KNOWHERE_INFO_ << "Set GPU pool size: init size " << init_size << ", max size " << max_size;
+    // TODO(wphicks): Is this actually used anywhere? If so, when in the
```
wphicks, line 630: Was not sure if this is being used at all or if we exclusively set the pool size by querying the environment variable.

If we actually use this somewhere, we should ensure that it happens before the `gpu_resources` object is initialized and then use the stored values for configuring the pool allocator on creation.

src/index/ivf_raft/ivf_raft_config.h
```diff
@@ -26,6 +26,7 @@ namespace knowhere {
 class RaftIvfFlatConfig : public IvfFlatConfig {
  public:
     CFG_LIST gpu_ids;
+    CFG_INT streams_per_device;
```
wphicks, line 29: Ideally, this would be a application-wide configuration rather than a per-index configuration, since only the first value will be used. I wasn't sure if there was a straightforward way to set this at the application level, though.

wphicks: Also, we should document this so that users understand the tradeoffs of setting this value higher. Higher number of streams per device may allow for higher throughput (depending on the size of the index and the batch size), but it will also lead to higher device memory usage, especially if the server is receiving many simultaneous requests.

src/index/ivf_raft/ivf_raft.cuh
Presburger: This api is used for init when milvus starts.

Presburger: milvus can parse the yaml configuration to set this option.

src/index/ivf_raft/ivf_raft_config.h
Presburger: This is a great job. However, the option option in config, is a json from the search task. streams_per_device should be a system configuration option, so my suggestion is that we can use environment variables or open a setup api.

Presburger: The original version was per-stream per-thread, and the simplification may also be an advantage.

src/index/cagra/cagra.cu
```diff
@@ -189,11 +168,9 @@ class CagraIndexNode : public IndexNode {
         is.read((char*)(&this->counts_), sizeof(this->counts_));
         this->devs_.resize(1);
         is.read((char*)(&this->devs_[0]), sizeof(this->devs_[0]));
-        auto scoped_device = detail::device_setter{devs_[0]};
+        auto scoped_device = raft_utils::device_setter{devs_[0]};
 
-        thread_local rmm::mr::cuda_memory_resource mr;
-        thread_local rmm::cuda_stream stream;
-        thread_local raft::device_resources res(stream.view(), nullptr, &mr);
+        auto& res = raft_utils::get_raft_resources();
```
Presburger, line 173: I would prefer not to use the explicit memory pool except for search. milvus is a distributed system, some nodes are only responsible for build index, they are not performance sensitive tasks to set up explicit pools, which sometimes cause deployment problems. For example, it is necessary to deploy build node and search node on the same card. Both processes will open two gpu memory pools at the same time, and if there is not enough gpu memory it will cause the deployment to fail.

wphicks: We already give users control over the starting and maximum pool size, and the default starting size is quite modest. If the separate pools grow to the point where an allocation fails, it would fail regardless of whether a pool is being used or not.

Basically, what I'm trying to say is that this is an issue regardless of whether or not a memory pool is used for search and index building or just search. In both situations, there's not enough memory, and there are other advantages to using a single unified memory allocator. In fact, it may make it easier to handle situations that push the available hardware to its memory limit by allowing for cleaner failures if we run out of memory.

src/index/ivf_raft/ivf_raft_config.h
wphicks: Great! Do we have a preference between using environment variables vs a setup API? We can handle the environment variable the exact same way you did the environment variable for pool size. If you'd rather use the setup API, though, could you point me to an example of how that's used elsewhere?

src/index/ivf_raft/ivf_raft.cuh
wphicks: Aha! Perfect. In that case, we can bring this back in. Any issue with letting it accept `std::optional<std::size_t>` instead of `size_t` to allow us to defer to the default rmm configuration if no option is provided?

achirkin:
Note, this may be partially relevant https://github.com/rapidsai/raft/pull/1356
With this, you should be able to easily set the memory limit for temporary (workspace) allocations in IVF-Flat / IVF-PQ (and also use a dedicated memory pool for performance).

src/index/cagra/cagra.cu
Presburger: @wphicks 
Thank you for your reply.
I can understand you your opinion. Often we are not very sensitive to build index performance. The problem with the memory pool is that once the memory inited or expands, it doesn't shrink back again. If it is a distributed scenario, buildnode and searchnode are two linux processes running on two separate containers, The two processes are two separate pools of gpu memory pool. it doesn't feel right to dedicate a section of gpu memory to a task that is not called very often.

Presburger:
@lowener Thanks for the pr, but remember to resolve the conflict code. 😊

Presburger:
> rapidsai/raft#1356

Great.

src/common/raft/raft_utils.cuh
```diff
@@ -0,0 +1,130 @@
...
+}
+
+inline void
+init_gpu_resources(std::size_t streams_per_device = std::size_t{1}, int device_id = get_current_device()) {
+    get_gpu_resources(streams_per_device).init(device_id);
+}
+
+inline auto&
+get_raft_resources(int device_id = get_current_device()) {
+    thread_local auto all_resources = std::map<int, std::unique_ptr<raft::device_resources>>{};
```
Presburger, line 119: why thread_local ?

src/index/cagra/cagra.cu
wphicks: Ah, I understand! How would we feel about allowing the user to optionally (e.g. via an environment variable) disable the memory pool altogether for a deployment? That would keep the case where search and build happen in the same process clean, but it would allow it to be disabled for the case of a dedicated build node. That might also better accommodate situations when search traffic is very "bursty" and we do not want to hang onto device memory corresponding to the peak.

src/common/raft/raft_utils.cuh
wphicks: This allows lazy instantiation of `device_resources` without requiring a lock every time we want to fetch the required `device_resources`. Individual `device_resources` may share the same stream (avoiding the problem this PR is intended to solve), but we can control how the streams and devices are distributed across threads.

Because we need to reference the correct `device_resources` very frequently, blocking all threads on that fetch would be expensive. Lazy instantiation is desirable because we may not know _a priori _ which host threads will need access to which devices.

wphicks:
Just took care of conflicts and fixed up something that got messed up in the merge. This should be ready for re-review.

Presburger (approved):

PR #966, opened 2023-06-28
Sync param check for DiskANN and HNSW

liliu-z:
This PR is to sync how HNSW&DiskANN will perform with empty config with Knowhere 1.x:

The rule is:
1. If `ef` exists, we need to ensure `ef` >= `k`.
2. If `ef` not exists, let `ef` = max(16, `k`).

### DIFF
--- a/src/index/hnsw/hnsw_config.h
+++ b/src/index/hnsw/hnsw_config.h
@@ -15,11 +15,15 @@
 #include "knowhere/comp/index_param.h"
 #include "knowhere/config.h"
 
-// This is not a valid EF value for HNSW
-// This value is used to tell if HnswConfig.ef is coming from user or not
-const int64_t kDefaultHnswEfPlaceholder = -1;
-
 namespace knowhere {
+
+namespace {
+
+constexpr const CFG_INT::value_type kEfMinValue = 16;
+constexpr const CFG_INT::value_type kDefaultRangeSearchEf = 16;
+
+}  // namespace
+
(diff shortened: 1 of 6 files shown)


src/index/hnsw/hnsw_config.h
```diff
@@ -15,11 +15,15 @@
...
 
-// This is not a valid EF value for HNSW
-// This value is used to tell if HnswConfig.ef is coming from user or not
-const int64_t kDefaultHnswEfPlaceholder = -1;
-
 namespace knowhere {
+
+namespace {
+
+constexpr const CFG_INT::value_type kDefaultEfDivider = 16;
```
cydrain, line 22: what "Divider" mean ?

src/index/hnsw/hnsw_config.h
```diff
@@ -15,11 +15,15 @@
...
+constexpr const CFG_INT::value_type kDefaultRangeSearchEf = 16;
```
cydrain, line 23: can we use one "kDefaultEf" for both search and range search ?

liliu-z: They don't have to be the same

src/index/hnsw/hnsw_config.h
liliu-z: meaning if k < divider, ef = divider. Anyway, more than welcome to better name, I agree it is poor

PR #974, opened 2023-06-28
Fix IVFFLATCC to support getVectors

foxspy:
#973


include/knowhere/config.h
```diff
@@ -427,8 +427,9 @@ class BaseConfig : public Config {
             .set_range(1, std::numeric_limits<CFG_INT::value_type>::max())
             .for_search();
         KNOWHERE_CONFIG_DECLARE_FIELD(num_build_thread)
-            .set_default(omp_get_max_threads())
             .description("index thread limit for build.")
+            .set_default(0)
+            .set_range(0, std::thread::hardware_concurrency())
```
cydrain, line 432: valid range should be (1, xxx)

CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ project(knowhere CXX C)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules/")
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR}/)
```
cydrain, line 20: why need this change ?

PR #983, opened 2023-07-05
Refactor binary metrics implementation

cydrain:
1. Rename files:
    BinaryDistance.h -> binary_distances.h
    BinaryDistance.cpp -> binary_distances.cpp
2. Raname APIs:
    binary_distance_knn_hc -> binary_knn_hc
    binary_distance_knn_mc -> binary_knn_mc
3. Replace SuperstructureComputer and SubstructureComputer with template StructureComputer<>
4. Add more unittests to cover metrics SUPERSTRUCTURE and SUBSTRUCTURE


thirdparty/faiss/faiss/utils/structure-inl.h
```diff
@@ -0,0 +1,416 @@
...
+        StructureComputer512() {}
+
+        StructureComputer512(const uint8_t *a8, int code_size) {
+            set(a8, code_size);
+        }
+
+        void set(const uint8_t *au8, int code_size) {
+            assert(code_size == 512);
+            const uint64_t *a = (uint64_t *)au8;
+            a0 = a[0]; a1 = a[1]; a2 = a[2]; a3 = a[3];
```
PwzXxm, line 274: why not use an array here?

thirdparty/hnswlib/hnswlib/hnswalg.h
```diff
@@ -1157,7 +1157,7 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
         }
 
         tableint currObj = enterpoint_node_;
-        auto vec_hash = knowhere::hash_vec((const float*)query_data, *(size_t*)dist_func_param_);
+        // auto vec_hash = knowhere::hash_vec((const float*)query_data, *(size_t*)dist_func_param_);
```
PwzXxm, line 1160: is this intentional?

cydrain: yes, Zihao will update knowhere::hash_vec() to support binary data later

thirdparty/faiss/faiss/utils/structure-inl.h
cydrain: since #1647, the code is like this authored by shengjun.
I think it's because accessing a0 is faster than accessing "a[0]"

PR #984, opened 2023-07-11
Update diskann benchmark

cydrain:


PwzXxm:
Also fixed setting AIO context pool.
/lgtm

Issue #514, opened 2022-10-27
Add ThreadPool to replace OMP


liliu-z:
OMP is hard to control and manage. It will amplify the number of threads, and its perform heavily depends on it implementation in different envs. 

We need to use a ThreadPool to replace it. At lease in the search path.

PR #989, opened 2023-07-12
Threadpool based HNSW insert

fhaltmayer:
Update the Add() of HNSW to use thread pool instead of using its own omp.


fhaltmayer:
https://github.com/milvus-io/knowhere/issues/514

cqy123456:
@fhaltmayer  Can you compare the hnsw indexing time before and after using the thread pool?

PR #998, opened 2023-07-21
fix folly compile problem

Presburger:


cydrain:
@hhy3 has fixed the compile issue in #1001

PR #1000, opened 2023-07-24
Fix HNSW range search result wrong

cydrain:


src/common/range_util.cc
```diff
@@ -59,8 +59,8 @@ GetRangeSearchResult(const faiss::RangeSearchResult& res, const bool is_ip, cons
                      const float range_filter, float*& distances, int64_t*& labels, size_t*& lims,
                      const BitsetView& bitset) {
     auto total_valid = CountValidRangeSearchResult(res, is_ip, nq, radius, range_filter, lims);
-    LOG_KNOWHERE_DEBUG_ << "Range search metric type: " << (is_ip ? "IP" : "L2") << ", radius " << radius
-                        << ", range_filter " << range_filter << ", total result num " << total_valid;
+    LOG_KNOWHERE_INFO_ << "Range search: is_ip " << (is_ip ? "True" : "False") << ", radius " << radius
+                       << ", range_filter " << range_filter << ", total result num " << total_valid;
```
liliu-z, line 63: Put `INFO` level log in the search path will be too heavy

cydrain: change back

PR #1014, opened 2023-07-28
Disable auto-remove for HNSW

yah01:
/kind improvement
Disable the file auto-remove for HNSW, we will manage the index file in Milvus


yah01:
The other PR has fixed this, close it

PR #1019, opened 2023-08-03
Fix brute force search for COSINE

cydrain:


src/common/comp/brute_force.cc
```diff
@@ -45,6 +43,14 @@ BruteForce::Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset
     auto xq = query_dataset->GetTensor();
     auto nq = query_dataset->GetRows();
 
+    bool is_cosine = IsMetricType(metric_str, metric::COSINE);
+    if (is_cosine) {
+        // must do copy here, otherwise it will affect GetVectorByIds
```
foxspy, line 48: May cause memory expanding. Chunk data may be copied multiple times (involving multiple growing segments and multiple search tasks)

PR #1021, opened 2023-08-04
Fix bruteforce cosine

hhy3:


src/common/comp/brute_force.cc
```diff
@@ -54,6 +50,13 @@ BruteForce::Search(const DataSetPtr base_dataset, const DataSetPtr query_dataset
     auto labels = new int64_t[nq * topk];
     auto distances = new float[nq * topk];
 
+    bool is_cosine = IsMetricType(metric_str, metric::COSINE);
+    std::unique_ptr<float[]> norms = nullptr;
+    if (is_cosine) {
+        norms = std::make_unique<float[]>(nb);
+        faiss::fvec_norms_L2(norms.get(), (const float*)xb, dim, nb);
```
yah01, line 57: Is it possible to calculate this from origin data directly? the xb could be much large and concurrent search requests will copy many data to run out of memory

liliu-z: To my understanding, we search growing chunk by chunk, which means the nb cannot be bigger than 1024. So 4KB memory is needed for each search.

The reason why put it here is to accelerate the calculation when nq is big

yah01: @liliu-z the sealed segment without index will also go brute force, and the sealed segment is with only 1 chunk, which with size of all rows

liliu-z: > @liliu-z the sealed segment without index will also go brute force, and the sealed segment is with only 1 chunk, which with size of all rows

Good catch!

liliu-z: @hhy3 Let's put it in

PR #1024, opened 2023-08-07
Remove info log of normalizing query while searching with cosine metric

chasingegg:
`Normalize` will be called at both `Train` and `Search` stage, which will generate many info logs while searching with cosine metric.

### DIFF
--- a/src/common/utils.cc
+++ b/src/common/utils.cc
@@ -50,7 +50,7 @@ Normalize(const DataSet& dataset) {
     auto dim = dataset.GetDim();
     float* data = (float*)dataset.GetTensor();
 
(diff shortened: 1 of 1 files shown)


src/common/utils.cc
```diff
@@ -50,8 +50,6 @@ Normalize(const DataSet& dataset) {
     auto dim = dataset.GetDim();
     float* data = (float*)dataset.GetTensor();
 
-    LOG_KNOWHERE_INFO_ << "vector normalize, rows " << rows << ", dim " << dim;
```
cydrain, line 53: you can change the log level from INFO to DEBUG

chasingegg: Done.

PR #1025, opened 2023-08-07
Fix cosine bruteforce

hhy3:


src/common/metric.h
```diff
@@ -27,7 +27,7 @@ Str2FaissMetricType(std::string metric) {
     static const std::unordered_map<std::string, faiss::MetricType> metric_map = {
         {metric::L2, faiss::MetricType::METRIC_L2},
         {metric::IP, faiss::MetricType::METRIC_INNER_PRODUCT},
-        {metric::COSINE, faiss::MetricType::METRIC_INNER_PRODUCT},
+        {metric::COSINE, faiss::MetricType::METRIC_COSINE},
```
chasingegg, line 30: Have you check if some parts of faiss rely on metric type, not only distance computation. Since it will cause some behavior changes, after this change knowhere's faiss indexes will use cosine natively

